PR #15, opened 2025-02-17
No need to initialize State directly.

mbrandonw:
Came across this while responding to #13. We are initializing a bunch of `State` values that weren't really necessary. Think this was just an artifact of us being deep in exploring how to use `@SharedReader` in SwiftUI views and we didn't audit it before releasing.

### DIFF
--- a/Examples/Reminders/ReminderForm.swift
+++ b/Examples/Reminders/ReminderForm.swift
@@ -12,31 +12,17 @@ struct ReminderFormView: View {
   @State var isPresentingTagsPopover = false
   @State var remindersList: RemindersList
   @State var reminder: Reminder
-  @State var selectedTags: [Tag]
+  @State var selectedTags: [Tag] = []
 
   @Dependency(\.defaultDatabase) private var database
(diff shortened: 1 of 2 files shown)


Examples/Reminders/ReminderForm.swift
```diff
@@ -12,31 +12,17 @@ struct ReminderFormView: View {
...
-    _remindersList = State(wrappedValue: remindersList)
-    if let existingReminder, let reminderID = existingReminder.id {
-      _reminder = State(wrappedValue: existingReminder)
-      do {
-        let tags = try _database.wrappedValue.read { db in
-          try Tag.all()
-            .joining(optional: Tag.hasMany(ReminderTag.self))
-            .filter(Column("reminderID").detached == reminderID)
-            .order(Column("name"))
-            .fetchAll(db)
```
mbrandonw, line 30: We were accidentally loading tags in the init of a view 😮. This has been moved to a `task` view modifier down below.

stephencelis (approved):
Looks good!

jan-martenn (approved):

PR #16, opened 2025-02-17
Update examples/docs to use `DatabasePool` for non-in-memory

stephencelis:
Database pools enable WAL mode and better multithreaded performance, so let's steer folks in that direction in the demo apps.

(I didn't update the case studies app since each study is very small and think we should do the minimal setup required.)

### DIFF
--- a/Examples/Reminders/ReminderForm.swift
+++ b/Examples/Reminders/ReminderForm.swift
@@ -225,9 +225,9 @@ struct TagsPopover: View {
 }
 
 #Preview {
-  let (remindersList, reminder) = prepareDependencies {
-    $0.defaultDatabase = .appDatabase
-    return try! $0.defaultDatabase.write { db in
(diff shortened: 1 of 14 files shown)


Examples/Reminders/ReminderForm.swift
```diff
@@ -225,9 +225,9 @@ struct TagsPopover: View {
 }
 
 #Preview {
-  let (remindersList, reminder) = prepareDependencies {
-    $0.defaultDatabase = .appDatabase
-    return try! $0.defaultDatabase.write { db in
+  let (remindersList, reminder) = try! prepareDependencies {
+    $0.defaultDatabase = try Reminders.appDatabase(inMemory: true)
```
stephencelis, line 229: I decided to preserve the `throws` since maybe folks would want to show an error with bug submission info upon launch instead of crash.

mbrandonw (approved):

PR #39, opened 2025-04-23
Disfavor an init to allow for bare @FetchAll syntax

mbrandonw:


Tests/SharingGRDBTests/FetchTests.swift
```diff
@@ -0,0 +1,53 @@
...
+import Sharing
+import SharingGRDB
+import StructuredQueries
+import SwiftUI
+import Testing
+
+@Suite(.dependency(\.defaultDatabase, try .database()))
+struct FetchTests {
+  @Test func bareFetchAll() async throws {
+    @FetchAll var records: [Record]
```
mbrandonw, line 13: Prior to this PR this kind of query would not actually fetch all records.

stephencelis (approved):

PR #41, opened 2025-04-29
fix linux build by importing GRDBSQLite

diederich:
I saw that the linux build failed and wondered if we should rely on the GRDBSQLite module instead of raw SQLite3.
From what I understand, that should resolve to SQLite3 on Darwin, but use libsqlite3 on Linux.

Also, I copied over the Linux pieces from [swift-sharing](https://github.com/pointfreeco/swift-sharing/blob/main/.github/workflows/ci.yml) & added installing sqlite.

Let me know what you think & happy to adjust!


stephencelis (approved):
Thanks!

PR #72, opened 2025-06-01
Modernize demo apps

mbrandonw:
We have a lot of big things coming to this library soon, so I wanted to take some time to modernize some of the demos a bit. Some of the things we have done:

* Upgraded IDs of tables to be UUIDs instead of auto incrementing integers. The former plays better with synchronizing across devices where it is quite important to have globally unique IDs.
* Moved logic of complex features (reminders lists, reminders detail and search) to their own `@Observable` model. This helped a great deal with animation glitches, testing and the views were becoming a nightmare to work with.
* Wrote some tests for the gnarly logic in a few features.
* Don't run any code in the entry point of the app when running in a preview context. Xcode is not capable of making a database connection to a file on disk during previews.
* Test databases are now persisted to disk in the temporary directory instead of run in-memory. This creates a more realistic environment for tests to run in a production environment.
* Changed database seeding so that it happens automatically in previews and provided an option in-app for seeding from the simulator. Tests also have their own seeding function so that seeds in previews/simulators are not conflated with previews.
* Controlled the database date functions so that we can better test queries.


Examples/Reminders/Schema.swift
```diff
@@ -41,10 +57,10 @@ extension Reminder {
 
 extension Reminder.TableColumns {
   var isPastDue: some QueryExpression<Bool> {
-    !isCompleted && #sql("coalesce(date(\(dueDate)) < date('now'), 0)")
+    !isCompleted && #sql("coalesce(date(\(dueDate)) < now(), 0)")
```
stephencelis, line 60: When [this](https://github.com/pointfreeco/swift-structured-queries/pull/61) lands we could potentially stay in our builder:

```suggestion
    !isCompleted && dueDate.map { $0 < Date() }
```

I know we like SQL to do as much work as possible, but maybe this is a case where it's more convenient and easier to control for tests by staying in Swift.

Examples/SyncUps/Schema.swift
```diff
@@ -4,22 +4,22 @@ import SwiftUI
 
 @Table
 struct SyncUp: Hashable, Identifiable {
-  let id: Int
+  let id: UUID
```
stephencelis, line 7: Should we leave other demos on the more common autoincrement IDs? Or should we plan to do more CloudKit in the demos?

Examples/SyncUps/SyncUpDetail.swift
```diff
@@ -140,27 +117,27 @@ struct SyncUpDetailView: View {
...
         }
       } header: {
         Text("Sync-up Info")
       }
 
-      if !model.details.meetings.isEmpty {
+      if !model.meetings.isEmpty {
         Section {
-          ForEach(model.details.meetings, id: \.id) { meeting in
+          ForEach(model.meetings, id: \.id) { meeting in
```
stephencelis, line 138:
```suggestion
          ForEach(model.meetings) { meeting in
```

Examples/SyncUps/SyncUpDetail.swift
```diff
@@ -178,7 +155,7 @@ struct SyncUpDetailView: View {
       }
 
       Section {
-        ForEach(model.details.attendees, id: \.id) { attendee in
+        ForEach(model.attendees, id: \.id) { attendee in
```
stephencelis, line 158:
```suggestion
        ForEach(model.attendees) { attendee in
```

PR #78, opened 2025-06-17
Fix FetchOne load() for optionals

rcarver:
Adds an overload for `FetchOne.load` where the value is optional. Previously this compiler error occurred:

image

I added the minimal to fix it, not sure if there are other cases!


stephencelis:
@rcarver Turns out we thought we shipped a bunch of `@FetchOne` fixes but they were on a branch. Can you take #79 for a spin and see if it fixes things?

stephencelis:
I believe this is fixed in #79, but let us know if we need to push things any further!

rcarver:
👍 @stephencelis thanks!

PR #92, opened 2025-07-06
fix #91

### LINKED ISSUE #91: Reminders Example `ReminderDetailModel.remindersQuery` incorrectly filters reminders
### Description

`RemindersDetailModel.remindersQuery` incorrectly filters reminders based on `!reminder.isCompleted` when `detailType` is `.all`.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] I have determined whether this bug is also reproducible in a vanilla GRDB project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/sharing-grdb/issues) or [discussion](https://github.com/pointfreeco/sharing-grdb/discussions).

### Expected behavior

Changing the value of `showCompleted` should return the appropriate set of reminders when `detailType == .all`: all reminders when `showCompleted == true` and all incomplete reminders when `showCompleted == false`.

### Actual behavior

When `detailType == .all` only incomplete reminders are returned regardless of the `showCompleted` value.

seanwoodward:
Fix for #91

### DIFF
--- a/Examples/Reminders/RemindersDetail.swift
+++ b/Examples/Reminders/RemindersDetail.swift
@@ -85,7 +85,7 @@ class RemindersDetailModel: HashableObject {
       .withTags
       .where { reminder, _, tag in
         switch detailType {
-        case .all: !reminder.isCompleted
+        case .all: if false { SQLQueryExpression("") } // break will not compile
         case .completed: reminder.isCompleted
         case .flagged: reminder.isFlagged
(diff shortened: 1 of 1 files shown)


seanwoodward:
@mbrandonw while the effect is the same, something about not emitting sql for that condition sits better with me. something to think about.

PR #104, opened 2025-07-19
Tests for GRDB+TaskLocal in async context

rcarver:
This adds tests to ensure GRDB behaves correctly in async contexts, and thus that `SharingGRDB` and related tools behave correctly.

* In general, do `TaskLocal` behave as expected inside async read/write
* Specifically, does `assertQuery` behave correctly in async read/write

The GRDB side in progress per this discussion: https://github.com/groue/GRDB.swift/discussions/1793

To verify the changes:

1. Run tests, they will fail on GRDB 7.5.0
2. Update GRDB to `development` branch and confirm tests pass


mbrandonw:
Hi @rcarver, thanks for taking the time to do this, but I'm not sure we need these tests in our library. They are only testing behavior in GRDB proper. It would also force us to update our dependence on GRDB to be the newest in order to guarantee tests pass, but I don't think it's necessary to force that update on everyone using this library.

I am going to close this for now, but if you have reason to believe that we should definitely have these tests in SharingGRDB we're open to discussing more!

rcarver:
@mbrandonw I found a new issue with GRDB+SnapshotTesting. It's technically different, but related enough I thought I'd keep the discussion here for now (all issues contribute to not being able to use `assertQuery` with GRDB)

### Description

Any use of `assertInlineSnapshot` within `database.read` (sync) in a test tagged `@MainActor` will [crash on this line of swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing/blob/b198a568ad24c5a22995c5ff0ecf9667634e860e/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift#L357)

This is unrelated the async issue, as I can confirm with GRDB 7.4.0 to 7.6.0

```swift
import GRDB
import InlineSnapshotTesting
import Testing

@MainActor
@Test
func snapshotInsideReadMainActor() throws {
  let database = try DatabaseQueue()
  try database.read { db in
    assertInlineSnapshot(of: "foo", as: .lines)  // 💥 crashes
  }
}
```

The same crash happens in XCTest without tagging `@MainActor` 

```swift
import GRDB
import InlineSnapshotTesting
import XCTest

final class SnapshotXCTests: XCTestCase {
  func testInlineSnapshotInsideDatabaseRead() throws {
    let database = try DatabaseQueue()
    try database.read { db in
      assertInlineSnapshot(of: "foo", as: .lines)  // 💥 crashes
    }
  }
}
```

You can also see the crash by simply adding `@MainActor` to this test from this PR https://github.com/pointfreeco/sharing-grdb/pull/104/files#diff-feea7bd1c079229efc7dc970c6c9fb9a114307d7952149f0b2d76dd75c7f6b04R28

---

Here is a full set of sync/async MainActor/non-MainActor tests just to fully understand what works and what doesn't.

```swift
import GRDB
import InlineSnapshotTesting
import Testing

@Test
func snapshotInsideRead() throws {
  let database = try DatabaseQueue()
  try database.read { db in
    assertInlineSnapshot(of: "foo", as: .lines) {
      """
      foo
      """
    }
  }
}

@MainActor
@Test("always fails")
func snapshotInsideReadMainActor() throws {
  let database = try DatabaseQueue()
  try database.read { db in
    assertInlineSnapshot(of: "foo", as: .lines)  // 💥 crashes
  }
}

@Test
func snapshotInsideAsyncRead() async throws {
  let database = try DatabaseQueue()
  try await database.read { db in
    assertInlineSnapshot(of: "foo", as: .lines) {
      """
      foo
      """
    }
  }
}

@MainActor
@Test("works with GRDB 7.6.0 only")
func snapshotInsideAsyncReadMainActor() async throws {
  let database = try DatabaseQueue()
  try await database.read { db in
    assertInlineSnapshot(of: "foo", as: .lines) {
      """
      foo
      """
    }
  }
}
```

rcarver:
@stephencelis works great!

* Confirmed this branch fixes the above unit tests
* Confirmed this branch fixed using `assertQuery` in my app's tests in a sync context
* Further confirmed that this branch + GRDB 7.6.0 fixes all known issues with async read/write, using dependencies and inline snapshot testing

PR #129, opened 2025-08-17
More sharing tests and improvements

mbrandonw:
With the improvements we made to the testability of sharing in #128 we can now write tests for behavior we added in #126. And it turns out we had a bug in that PR 😬. We are preventing changes being sync'd from CK to the user device when they have read only permissions. We need to bypass permission checks when rows are updated from the sync engine directly.

### DIFF
--- a/Sources/SharingGRDBCore/CloudKit/CloudKitSharing.swift
+++ b/Sources/SharingGRDBCore/CloudKit/CloudKitSharing.swift
@@ -32,6 +32,7 @@ extension SyncEngine {
       case recordNotRoot([ForeignKey])
       case recordTableNotSynchronized
       case recordTablePrivate
+      case shareNotFound
     }
 
     let recordTableName: String
@@ -40,7 +41,7 @@ extension SyncEngine {
     let debugDescription: String
 
     var errorDescription: String? {
-      "The record could not be shared."
+      "An error occured when editing sharing."
     }
   }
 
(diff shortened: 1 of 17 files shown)


Sources/SharingGRDBCore/CloudKit/CloudKitSharing.swift
```diff
@@ -40,7 +41,7 @@ extension SyncEngine {
     let debugDescription: String
 
     var errorDescription: String? {
-      "The record could not be shared."
+      "An error occured when editing sharing."
```
mbrandonw, line 44: Made this a bit more generic because we have a new error that can be emitted when you try to un-share a record that has no associated share.

Sources/SharingGRDBCore/CloudKit/Triggers.swift
```diff
@@ -238,7 +238,8 @@
           )
         }
         .where {
-          $0.parentRecordName.is(nil)
+          !SyncEngine.isSynchronizingChanges()
```
mbrandonw, line 241: This is the bug we had. We need to check permissions only when not updating from the sync engine.

Tests/SharingGRDBTests/CloudKitTests/AccountLifecycleTests.swift
```diff
@@ -59,7 +59,7 @@ extension BaseCloudKitTests {
 
       try await syncEngine.processPendingDatabaseChanges(scope: .private)
       try await syncEngine.processPendingRecordZoneChanges(scope: .private)
-      assertInlineSnapshot(of: syncEngine.container, as: .customDump) {
+      assertInlineSnapshot(of: container, as: .customDump) {
```
mbrandonw, line 62: I made this change across the entire test suite so it's a little noisy. but we have direct access to `container` in the tests so no need to go through the sync engine.

PR #132, opened 2025-08-20
StructuredQueries 0.13.0 Support

stephencelis:
There has been a small breaking change to the `Seeds` API that we need to address in StructuredQueries.


thimic:
Is this change needed on the `cloudkit` branch too? I updated packages today and got a build error when using seed. Pinning StructuredQueries to `0.12.1` got things going again.

PR #133, opened 2025-08-21
SharingGRDBTestSupport with assertQuery

rcarver:
- **create SharingGRDBTestSupport with assertQuery from StructuredQueries**
- **rework the api to assertQuery(includeSQL)**

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -20,6 +20,10 @@ let package = Package(
       targets: ["SharingGRDBCore"]
     ),
     .library(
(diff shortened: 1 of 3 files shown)


stephencelis:
@rcarver This looks good! But we got a test failure for `assertQuery` in release since we don't pretty-print queries in release mode. Want to wrap those tests in `#if DEBUG`?

stephencelis (approved):
Looks great, thanks!

PR #135, opened 2025-08-21
Improve some docs.

mbrandonw:


Sources/SharingGRDBCore/Documentation.docc/Articles/PreparingDatabase.md
```diff
@@ -221,6 +218,39 @@ database connection:
...
+}
+```
+
+It may seem counterintuitive that we recommend using SQL strings for table definitions when so much
+of the library provides type-safe and schema-safe tools for executing SQL. But table definition SQL
+is fundamentally different from other SQL. Read [this article] from our StructuredQueries library
+to learn more about this decision.
+
+[this article]: https://swiftpackageindex.com/pointfreeco/swift-structured-queries/main/documentation/structuredqueriescore/definingyourschema#Table-definition-tools
+
```
7andres, line 253: I think it would be important to find a way to link the information at the top of this section about grdb-migration-docs—which also explains how to perform migrations using Swift code—with this StructuredQueries way

It would also be good to highlight that there are several ways to complete the code snippet marked with // Execute SQL to create tables, and it’d be helpful to include links [or clarify] both to the GRDB approach and to the StructuredQueries approach using the #sql macro!

7andres (approved):

PR #144, opened 2025-08-29
CloudKit: Compile tests on non-macOS platforms

stephencelis:
Should hopefully get us closer to green CI.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,7 +36,7 @@ let package = Package(
     .default(enabledTraits: ["SharingGRDBTagged"]),
   ],
   dependencies: [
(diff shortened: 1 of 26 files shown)


Package.swift
```diff
@@ -36,7 +36,7 @@ let package = Package(
     .default(enabledTraits: ["SharingGRDBTagged"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/groue/GRDB.swift", from: "7.4.0"),
+    .package(url: "https://github.com/groue/GRDB.swift", from: "7.6.0"),
```
stephencelis, line 39: Given the escaping closure fix in 7.6 I think it's a good idea to bump this...

mbrandonw (approved):

PR #151, opened 2025-09-02
Delete CK record when updating primary key.

mbrandonw:
If we detect that someone has changed the primary key of a row (which should be rare, but can happen), we need to be sure to delete the record in CloudKit with the old primary key.

### DIFF
--- a/Tests/SharingGRDBTests/CloudKitTests/FetchRecordZoneChangesTests.swift
+++ b/Tests/SharingGRDBTests/CloudKitTests/FetchRecordZoneChangesTests.swift
@@ -481,119 +481,14 @@ extension BaseCloudKitTests {
       }
       try await syncEngine.processPendingRecordZoneChanges(scope: .private)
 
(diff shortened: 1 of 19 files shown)


Tests/SharingGRDBTests/CloudKitTests/FetchRecordZoneChangesTests.swift
```diff
@@ -480,14 +481,24 @@ extension BaseCloudKitTests {
       }
       try await syncEngine.processPendingRecordZoneChanges(scope: .private)
 
+      assertQuery(SyncMetadata.select(\.recordName), database: userDatabase.database) {
+        """
+        ┌────────────────────┐
+        │ "1:remindersLists" │
+        │ "1:reminders"      │
+        │ "1:reminderTags"   │
+        │ "optional:tags"    │
```
mbrandonw, line 490: This shows that the "weekend" tag was deleted from the sync metadata.

PR #154, opened 2025-09-03
More docs

mbrandonw:
Getting some basics docs on most public APIs.


Sources/SQLiteData/CloudKit/CloudKitSharing.swift
```diff
@@ -116,7 +139,6 @@
             return try await container.database(for: rootRecord.recordID)
               .record(for: shareRecordID) as? CKShare
           } catch let error as CKError where error.code == .unknownItem {
-            reportIssue("This would have been a problem before")
```
mbrandonw, line 119: can't remember what this report issue is about, but doesnt seem very helpful.

PR #163, opened 2025-09-06
More efficient joins to SyncMetadata table

mbrandonw:
Right now we recommend people join their tables to `SyncMetadata` using the `recordName` helper we defined on PKT, and under the hood it uses `_recordName` which just concatenates primary key and record type in SQL. That unfortunately prevents SQLite from being able to use any indices, and so such queries will be slower than necessary.

I recommend getting rid of the `recordName` helper, and instead have a helper that allows people to join on the `primaryKey` _and_ `recordType` at the same time (note it is not possible to join only on primary key because multiple rows in `SyncMetadata` can have the same `recordPrimaryKey`, such is the case for `RemindersList` and `RemindersListAsset`).

There could also be the possibility of some kind of `T.withSyncMetadata`, but wasn't able to juggle the generics just yet.


Examples/Reminders/RemindersLists.swift
```diff
@@ -15,7 +15,9 @@ class RemindersListsModel {
       .leftJoin(Reminder.all) {
         $0.id.eq($1.remindersListID) && !$1.isCompleted
       }
-      .leftJoin(SyncMetadata.all) { $0.recordName.eq($2.recordName) }
+      .leftJoin(SyncMetadata.all) {
+        #sql("\($0.recordTypeAndPrimaryKey) = \($2.recordTypeAndPrimaryKey)")
```
mbrandonw, line 19: Ideally we could do:

```
$0.recordTypeAndPrimaryKey.eq($2.recordTypeAndPrimaryKey)
```

…but `eq` currently only works for `QueryBindable` type (which tuple is not).

We can also consider different ways of expressing this condition. For example, we already have `PKT.metadata(for:)` for constructing a `Where` for the metadata that belongs to a particular primary key, but that's not really reusable here.

Sources/SQLiteData/CloudKit/SyncMetadata.swift
```diff
@@ -163,8 +170,18 @@
 
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
   extension PrimaryKeyedTableDefinition where PrimaryKey.QueryOutput: IdentifierStringConvertible {
-    public var recordName: some QueryExpression<String> {
-      _recordName
```
mbrandonw, line 167: I don't think we should publicly expose this at all, it's too dangerous. We still have the internal `_recordName` one which we use for selects, which is fine.

PR #165, opened 2025-09-07
Fix tests for iOS.

mbrandonw:
Tests aren't currently building for iOS, so this fixes that.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -251,8 +251,6 @@
       try validateSchema()
     }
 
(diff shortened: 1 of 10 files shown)


Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -251,8 +251,6 @@
       try validateSchema()
     }
 
-    @TaskLocal package static var _isSynchronizingChanges = false
```
mbrandonw, line 254: I moved this out of the `SyncEngine` so that it didn't have the available restrictions. This made it annoying to use `userDatabase.write` from inside a test traits.

PR #188, opened 2025-09-16
Support moving records across zones

mbrandonw:
Currently we do not support moving records across zones (for example, you cannot remove a reminder from a shared list to a non-shared list, or vice-versa). When we detect this we throw an error (see #179).

Well, now that we have #187 we can properly detect zones at the `SyncMetadata` level, and that makes it easier to support this. When we detect a record is moving zones we gather all of its child records (and grandchild, etc) and delete all of those records from CloudKit in the old zone, and then re-save them in the new zone. All while keeping the local data intact so the user shouldn't notice anything happening behind the scenes.

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
@@ -202,15 +202,8 @@
               zoneName: new.zoneName,
               ownerName: new.ownerName,
               oldZoneName: new.zoneName,
-              oldOwnerName: new.ownerName
-                //              lastKnownServerRecord: new.lastKnownServerRecord
-                //                ?? rootServerRecord(recordName: new.recordName),
-                //              newParentLastKnownServerRecord: parentLastKnownServerRecord(
-                //                parentRecordPrimaryKey: new.parentRecordPrimaryKey,
-                //                parentRecordType: new.parentRecordType
-                //              ),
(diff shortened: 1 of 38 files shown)


Sources/SQLiteData/CloudKit/Internal/Triggers.swift
```diff
@@ -214,19 +217,48 @@
...
+              oldOwnerName: old.ownerName,
+              childRecordNames: #sql(
+                """
+                iif(
+                  \(zoneChanged),
+                  \(descendantRecordNamesJSON),
+                  NULL
+                )
+                """
+              )
```
stephencelis, line 261: Could this be the following?
```suggestion
              childRecordNames: Case().when(zoneChanged, then: descendantRecordNamesJSON)
                """
              )
```

Sources/SQLiteData/CloudKit/Internal/Triggers.swift
```diff
@@ -133,7 +138,8 @@
 
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
   extension SyncMetadata {
-    fileprivate static func upsert<T: PrimaryKeyedTable, Name>(
+
```
stephencelis, line 141:
```suggestion
```

stephencelis (approved):

PR #190, opened 2025-09-16
UIApplication is not available on watchOS

thimic:
Only reference `UIApplication` on iOS, watchOS fails to build otherwise.

Possibly too simplistic, didn't have a lot of time to look at it.


mbrandonw (approved):
Good catch @thimic, thank you!

PR #191, opened 2025-09-16
Fix assertQuery changing to empty results after snapshot contains results

rcarver:
Fixes the issue brought up in slack

image


stephencelis:
@rcarver Can you give an example of what this fixes? Maybe a test with an expected failure or something to that effect?

I think the fix we probably want to apply here is to always record an empty result set, something like:

```swift
assertQuery(Reminder.all) {
  """
  (No results.)
  """
}
```

Think you could cook that change up?

rcarver:
Also clarified the 'no results' case

```swift
  @Test func assertQueryEmpty() throws {
    assertQuery(
      Record.all.where { $0.id == -1 }.select(\.id)
    ) {
      """
      (No results)
      """
    }
  }
```

mbrandonw:
I just merged main to get everything caught up and re-recorded the SQLiteData tests which now properly shows "(no results)" in a bunch of tests.

rcarver:
Thanks! lgtm

PR #217, opened 2025-09-23
Support lazy database initialization

kabiroberai:
This PR adds a `DependencyValues.getDefaultDatabase` which can be set to a closure that's invoked each time `defaultDatabase` is accessed. Further, `defaultDatabase` is updated to use `getDefaultDatabase` under the hood.

This enables a couple of use cases:

### Lazy Initialization

Having to (generally) set `defaultDatabase` in `prepareDependencies` means that you currently have to initialize the DB very early in the app's lifecycle. By using `getDefaultDatabase` instead, it's possible to split up the "configuration" (`prepareDependencies`) and "initialization" (first access of `defaultDatabase`) into separate steps. This is especially useful if the DB isn't needed until some point later in the app's lifecycle.

```swift
@main
struct MyApp: App {
  // Lazily creates database connection and run migrations on first use
  private static let database = try! appDatabase()

  init() {
    prepareDependencies {
      $0.getDefaultDatabase = { Self.database }
    }
  }
  // ...
}
```

### Side Effects

In some cases, you might want to warn/error-out if something tries to access the DB. For example, if you only support the database in your main app target and want to `fatalError` if any code in an app extension target attempts to access the database. Currently, one way to do this is to simply not assign `defaultDatabase` at all, at which point accessing it will trigger `reportIssue`, which can forward to a custom issue reporter if you so wish. But that's a pretty wide net (there can be other issues you don't want to `fatalError` on). Instead, it's much nicer to be able to do something like

```swift
prepareDependencies {
  if isMainTarget {
    $0.defaultDatabase = try! appDatabase()
  } else {
    $0.getDefaultDatabase = { fatalError("Cannot access DB from an App Extension") }
  }
}
```

### DIFF
--- a/Sources/SQLiteData/StructuredQueries+GRDB/DefaultDatabase.swift
+++ b/Sources/SQLiteData/StructuredQueries+GRDB/DefaultDatabase.swift
@@ -82,13 +82,44 @@ extension DependencyValues {
   ///
   /// See <doc:PreparingDatabase> for more info.
   public var defaultDatabase: any DatabaseWriter {
+    get { self.getDefaultDatabase() }
+    set { self.getDefaultDatabase = { newValue } }
+  }
+
+  /// A function that returns the default database used by `fetchAll`, `fetchOne`, and `fetch`.
+  ///
+  /// You can use this instead of ``defaultDatabase`` to defer creating the database until
+  /// it is first needed, and/or to perform side effects when the database is accessed.
+  /// For example:
+  ///
+  /// ```swift
+  /// import SQLiteData
+  /// import SwiftUI
+  ///
+  /// @main
+  /// struct MyApp: App {
+  ///   // Lazily create database connection and run migrations...
+  ///   private static let database = try! DatabaseQueue(/* ... */)
+  ///
+  ///   init() {
+  ///     prepareDependencies {
+  ///       $0.getDefaultDatabase = { Self.database }
+  ///     }
+  ///   }
+  ///   // ...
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @kabiroberai! Thanks for taking the time to PR. We're not opposed to the change, but we can't say we understand the motivation yet. Can you provide a bit more information about why you need such a feature with some real world examples?

kabiroberai:
> You can always run migrations after the database connection has been provisioned. It doesn't need to be apart of the initial creation of the database connection.

Oh I somehow missed that migrations don't have to be run on setup. That does solve that.

> A custom issue reporter (like you mentioned) would work because you could inspect the message to determine whether or not to report it. Makes it possible to report/fatal error on only those errors.

"you could inspect the message to determine ..." is exactly what I'm trying to avoid :/ is there a way for the error to not be stringly-typed?

mbrandonw:
> "you could inspect the message to determine ..." is exactly what I'm trying to avoid :/ is there a way for the error to not be stringly-typed?

That would be a pretty big project to undertake. I can imagine there are a few ways to approach the problem, and not sure if we will easily come to the best one without more consideration or more real world examples.

But I do think it's pretty safe to say that if the message contains the phrase "defaultDatabase", then it is probably the issue you want to report.

kabiroberai:
Closing after discussing this on Slack: we're going to explore a more general "pass a factory to `prepareDependencies`" strategy on the side.

PR #226, opened 2025-09-26
SyncEngine primary key migration tool

stephencelis:
CloudKit synchronization requires tables with a particular design: they must have a primary key that uniquely identifies the record (this includes join tables).

It is quite common for SQLite tables to adopt auto-incrementing integer primary keys by default, which are not compatible with synchronization, so to adopt CloudKit many users will need to migrate their databases to these constraints. While [the steps](https://swiftpackageindex.com/pointfreeco/sqlite-data/main/documentation/sqlitedata/cloudkit#Preparing-an-existing-schema-for-synchronization) are somewhat straightforward, there are many of them, much code to write, and many opportunities to get something wrong.

Luckily, we can automate the common cases, and this PR provides a tool for just that:

```swift
migrator.registerMigration("Migrate primary keys to UUID") { db in
  try SyncEngine.migratePrimaryKeys(
    db,
    tables: RemindersList.self,
    Reminder.self
    ReminderTag.self,
    Tag.self
  )
}
```


diatoming:
Thanks for this primary keys migration tool, very helpful.
My old database schema with Int64 as id fixed without problem, but unfortunately there are some unique constraints added to some columns too.
Since the SyncEngine does now allow such constraints, can we add an option to also allow removing unique constraint at the same time when converting to UUID with method 'SyncEngine.migratePrimaryKeys'?

mbrandonw:
@diatoming sorry I misspoke. I thought unique columns were secretly powered by a unique index in SQLite, and so I thought you could just drop that index. But that is not true, and so I suppose we would need to add an option to drop uniqueness constraints. Would you have any interest in tackling that?

PR #245, opened 2025-10-07
Compute statement hashability from underlying query fragment

stephencelis:
We currently depend directly on the hashability of `Statement` in the various fetch property wrappers, but this conformance has been found to be problematic and so we should remove it. We can make this change today, though, before removing it, by computing the hashability of a statement from its query fragment data.


Examples/Reminders/SearchReminders.swift
```diff
@@ -178,6 +178,14 @@ class SearchRemindersModel {
...
       )
     }
+    static func == (lhs: Self, rhs: Self) -> Bool {
+      lhs.baseQuery.query == rhs.baseQuery.query
+        && lhs.showCompletedInSearchResults == rhs.showCompletedInSearchResults
+    }
+    func hash(into hasher: inout Hasher) {
+      hasher.combine(baseQuery.query)
+      hasher.combine(showCompletedInSearchResults)
+    }
```
stephencelis, line 188: This is required when we make the change in StructuredQueries...so maybe we need to think on this more.

mbrandonw (approved):

PR #246, opened 2025-10-08
Don't execute `@FetchOne` queries for `_Selection`s.

stephencelis:
We added an overload to handle `@FetchAll`, but we missed this one.


stephencelis:
Depends on https://github.com/pointfreeco/swift-structured-queries/pull/204

mbrandonw (approved):

PR #259, opened 2025-10-15
Fixes with sharing records

mbrandonw:
Right now we don't cache the newest server records when creating a share, which means if you share again we may use old records with old `recordChangeTag`s. This PR now refreshes server records whenever sharing a record.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -358,7 +358,7 @@
     public func start() async throws {
       try await start().value
     }
-    
+
     /// Determines if the sync engine is currently sending local changes to the CloudKit server.
     ///
     /// It is an observable value, which means if it is accessed in a SwiftUI view, or some other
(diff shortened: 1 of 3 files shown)


Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -872,19 +872,31 @@
         )
 
       case .willFetchRecordZoneChanges:
-        fetchingChangesCount += 1
+        await MainActor.run {
+          fetchingChangesCount += 1
+        }
```
mbrandonw: Unrelated to this PR but something I noticed. If we don't update observable properties on the main thread then SwiftUI may get stuck with old data.

stephencelis (approved):

Tests/SQLiteDataTests/CloudKitTests/SharingTests.swift
```diff
@@ -644,6 +644,65 @@
...
+          MockCloudContainer(
+            privateCloudDatabase: MockCloudDatabase(
+              databaseScope: .private,
+              storage: [
+                [0]: CKRecord(
+                  recordID: CKRecord.ID(share-1:remindersLists/zone/__defaultOwner__),
+                  recordType: "cloudkit.share",
+                  parent: nil,
+                  share: nil,
+                  cloudkit.title: "Please join my list!"
```
mbrandonw: This would fail previously because the second time sharing the record did not actually work.

PR #270, opened 2025-10-22
Use metadatabase to query sync metadata.

mbrandonw:
All reads/writes to the metadatabase should happen through the dedicated metadatabase connection because the user's connection may not have the metadatabase attached. This can be tricky to get right, and we seem to have a spot where we forgot to use the metadatabase directly.

- [ ] Add a test that deletes shares when the metadatabase is not attached.

### DIFF
--- a/Tests/SQLiteDataTests/CloudKitTests/AttachedMetadatabaseTests.swift
+++ b/Tests/SQLiteDataTests/CloudKitTests/AttachedMetadatabaseTests.swift
@@ -1,0 +1,69 @@
+#if canImport(CloudKit)
+  import CloudKit
+  import ConcurrencyExtras
+  import CustomDump
+  import InlineSnapshotTesting
+  import OrderedCollections
+  import SQLiteData
+  import SQLiteDataTestSupport
+  import SnapshotTestingCustomDump
+  import Testing
+
+  extension BaseCloudKitTests {
+    @MainActor
+    @Suite(.attachMetadatabase(true))
+    final class AttachedMetadatabaseTests: BaseCloudKitTests, @unchecked Sendable {
+      @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+      @Test func basics() async throws {
+        let remindersList = RemindersList(id: 1, title: "Personal")
+        try await userDatabase.userWrite { db in
+          try db.seed {
+            remindersList
+          }
+        }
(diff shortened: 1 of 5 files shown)


stephencelis (approved):

Tests/SQLiteDataTests/CloudKitTests/AttachedMetadatabaseTests.swift
```diff
@@ -0,0 +1,69 @@
+#if canImport(CloudKit)
```
mbrandonw: Added a new test suite to specifically test joining to the metadatabase when it is attached.

Tests/SQLiteDataTests/CloudKitTests/SharingTests.swift
```diff
@@ -2753,6 +2753,52 @@
           """
         }
       }
+
+      @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+      @Test func deleteShare() async throws {
```
mbrandonw: This test was failing prior to the change made in this PR.

Tests/SQLiteDataTests/Internal/BaseCloudKitTests.swift
```diff
@@ -11,7 +11,8 @@ import os
   .dependencies {
     $0.currentTime.now = 0
     $0.dataManager = InMemoryDataManager()
-  }
+  },
+  .attachMetadatabase(false)
```
mbrandonw: Let's not attach the metadatabase in tests by default, and instead opt into it explicitly.

Tests/SQLiteDataTests/Internal/CloudKitTestHelpers.swift
```diff
@@ -20,9 +20,10 @@ extension PrimaryKeyedTable where PrimaryKey.QueryOutput: IdentifierStringConver
 
 @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
 extension SyncEngine {
-  struct ModifyRecordsCallback {
-    fileprivate let operation: @Sendable () async -> Void
-    func notify() async {
+  struct ModifyRecordsCallback<ReturnValue> {
```
mbrandonw: Currently when we invoke `notify()` in tests it doesn't return the results from the `modify…` call. So I updated this bit of library code to allow for that in future tests that may want to assert on that information directly.

PR #288, opened 2025-11-01
Make SharedRecord CKContainer public

larryonoff:
Discussed in https://github.com/pointfreeco/sqlite-data/pull/280


mbrandonw:
Hi @larryonoff, as far as I can tell the two main things you need are the existing share and the `CKContainer`. The existing share can be found by querying the `SyncMetadata` table, and the `CKContainer` should already be known to you because you are the one who configured your project. And if you are making library code then you can make it so that the user of your library has to pass the explicit `CKContainer` since then they would be the ones to know what the container is.

PR #289, opened 2025-11-01
Add existingShare

larryonoff:
The method fetches existing `ShareRecord` without creating it.

Use case: `SWCollaborationView` should show only when `CKShare` exists, when `SyncEngine.share(record:configure:)` creates `CKShare` if it doesn’t exist.

### DIFF
--- a/Sources/SQLiteData/CloudKit/CloudKitSharing.swift
+++ b/Sources/SQLiteData/CloudKit/CloudKitSharing.swift
@@ -208,6 +208,17 @@
       return SharedRecord(container: container, share: savedShare)
     }
 
+    public func existingShare<T: PrimaryKeyedTable>(record: T) async throws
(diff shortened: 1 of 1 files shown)


stephencelis:
@larryonoff I think we'd like to lean on the existing querying tools where we can for now before adding more helpers.

Also, you should prefer finding using the `syncMetadataID`, instead: 

```swift
try SyncMetadata
  .find(reminder.syncMetadataID)
  .select(\.share)
  .fetchOne(db) ?? nil
```

The `recordName` API is not the most efficient way of querying, and we should probably figure out if it's even necessary these days, or if we should deprecate it.

The above should hopefully not be _too_ much of a mouthful for folks as opposed to the helper you define in this PR, but we may consider a helper in the future if this comes up again.

PR #302, opened 2025-11-21
Data backing

johankool:
Regarding [this message on Slack](https://pointfreecommunity.slack.com/archives/C08CZUG4Z9D/p1763137137826489) and [this discussion](https://github.com/pointfreeco/sqlite-data/discussions/281): could this be an approach you would consider to be able to customize wether a column is stored as `Data` in CloudKit or as a `CKAsset`?


mbrandonw:
Hi @johankool, thanks for exploring this. We will chat about this approach, but by far the most dangerous part as it stands right now is that if you switch a column from `.inline` to `.asset` (or vice versa) after having deployed your app, then the next update of your app will crash. `CKRecord`s emit an immediate crash if you try to set a key in both the unencrypted and encrypted values.

stephencelis:
@johankool We chatted about this today and still think the automatic approach mentioned in https://github.com/pointfreeco/sqlite-data/discussions/281 is the way to go. We could start with an arbitrary, hard-coded threshold (10kb? 16?) and then could explore further configuration/behavior in the future.

This approach would hopefully minimize the additional API surface area and folks could benefit with minimal configuration.

Are you open to updating this PR with this alternate approach?

Tests/SQLiteDataTests/CloudKitTests/AssetsTests.swift
```diff
@@ -320,12 +510,12 @@
...
-          ┌─────────────────────────────┐
-          │ RemindersListAsset(         │
-          │   remindersListID: 1,       │
-          │   coverImage: Data(5 bytes) │
-          │ )                           │
-          └─────────────────────────────┘
+          ┌──────────────────────────────────┐
+          │ RemindersListAsset(              │
+          │   remindersListID: 1,            │
+          │   coverImage: Data(16,811 bytes) │
```
johankool: Need to fix this in CustomDump: on machines with a thousands separator that is not a comma this test fails.

johankool: A fix is proposed here: https://github.com/pointfreeco/swift-custom-dump/pull/163

PR #310, opened 2025-11-25
Private tables should always be saved in private db.

mbrandonw:
We currently have a bug with private tables + sharing where we accidentally try to save rows from the private table in the shared db, and that causes a permission failure. This PR makes sure that private tables are always saved in the private DB.

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
@@ -6,20 +6,38 @@
   extension PrimaryKeyedTable {
     static func metadataTriggers(
       parentForeignKey: ForeignKey?,
-      defaultZone: CKRecordZone
+      defaultZone: CKRecordZone,
+      privateTables: [any SynchronizableTable]
     ) -> [TemporaryTrigger<Self>] {
       [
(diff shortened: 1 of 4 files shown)


Sources/SQLiteData/CloudKit/Internal/Triggers.swift
```diff
@@ -322,15 +354,24 @@
...
-      parentForeignKey
+    let zoneNameOverride: SQLQueryExpression<String?>
+    let ownerNameOverride: SQLQueryExpression<String?>
+    if privateTables.contains(where: { $0.base.tableName == Base.tableName }) {
+      zoneNameOverride = #sql("\(quote: defaultZone.zoneID.zoneName, delimiter: .text)")
+      ownerNameOverride = #sql("\(quote: defaultZone.zoneID.ownerName, delimiter: .text)")
+    } else {
+      zoneNameOverride = #sql("NULL")
+      ownerNameOverride = #sql("NULL")
+    }
```
mbrandonw: This is the main fix. If we are getting parent information for a private table we can make sure the zone is the default zone. We could probably clean this up a bit, I mostly just focused on getting a passing test.

stephencelis (approved):

PR #314, opened 2025-11-28
Update permission check for CloudKit sharing

johankool:
Fix code snippet using non-existing permission property


mbrandonw (approved):
Thanks!

PR #317, opened 2025-12-02
Add indexes on foreign keys in Reminders and SyncUp examples

jcmourey:
Adds index to foreign-key columns in the SQLiteData Reminders and SyncUp examples, as Brandon requested in Slack. Good practice to help query performance.


mbrandonw (approved):
Thanks @jcmourey!

PR #322, opened 2025-12-05
Fix id column type in documentation example schema to prevent error

RobbieClarken:
When the `id` column is defined as `INT NOT NULL PRIMARY KEY AUTOINCREMENT` sqlite will throw an error:

> AUTOINCREMENT is only allowed on an INTEGER PRIMARY KEY

This is fixed by changing `INT` to `INTEGER`. I also changed the other columns for consistency.

The error was being thrown for sqlite-data 1.4.0, iOS 26.1.


mbrandonw (approved):
Hi @RobbieClarken, good catch!

PR #323, opened 2025-12-09
Fix malformed link to GRDB in docs

lukaskubanek:
This PR fixes a malformed link to GRDB in the documentation (see [the first paragraph](https://swiftpackageindex.com/pointfreeco/sqlite-data/1.4.0/documentation/sqlitedata)). And while at it, it inlines all links to GRDB in the Markdown to unify the style with other links.


mbrandonw (approved):
Good catch @lukaskubanek!

It's a little annoying that that style of markdown link does not work across `@` blocks (like `@Row`). I guess docc processes all the text between the `@` blocks as individual markdown units and so links cannot be shared across them. Might be worth a bug report on docc even.

PR #335, opened 2025-12-18
Handle 'quotaExceeded' errors

mbrandonw:
If the user runs out of iCloud storage, saving records will report a `.quotaExceeded` error. Currently we are not handling that error correctly. Those updates will never be retried again later (unless the record happens to be edited later). And there is currently no way for the user of our library to be notified of such errors so that they can let the user know that there is something for them to fix. This PR addresses both of those problems:

* When we detect a `.quotaExceeded` error we not save the record ID to the `PendingRecordZoneChanges` table, which means they will be re-enqueued whenever the sync engine starts again. This will give the records a chance to save again at a later time.
  * Related to this, we also now retry those records if the `sendChanges` method is every explicitly called on the sync engine.
* Added a new delegate method that allows listening for when saving records caused a `.quoteExceeded` error.


Sources/SQLiteData/CloudKit/SyncEngineDelegate.swift
```diff
@@ -78,6 +78,32 @@
...
+    ///
+    /// - Parameters:
+    ///   - syncEngine: The sync engine that generates the event.
+    ///   - quotaExceeded: Determines if records failed to save due to a 'quotaExceeded` error.
+    ///   - scope: The database that the event occured on.
+    func syncEngine(
+      _ syncEngine: SyncEngine,
+      quotaExceeded: Bool,
+      scope: CKDatabase.Scope
+    )
```
mbrandonw: Should we surface the record/participant/CKShare that is responsible for the error?

PR #350, opened 2025-12-31
Don't execute `@Fetch{All,One}` queries for selections

stephencelis:
It's currently possible to execute these queries if you specify a parameter like `animation`. These parameters aren't actually useful for a selection query, so let's deprecate those overloads.


NataliaMihailiuc (changes requested):
Kjhgjj

mbrandonw (approved):

PR #359, opened 2026-01-06
Fix 'limitExceeded' errors related to FK constraint failures.

### LINKED ISSUE #333: CloudKit sync fails with "Limit Exceeded" error when syncing more than 400 items
### Description

When synchronizing tables with a large number of records (e.g., LocationPoint with GPS coordinates), CloudKit sync fails with a "Limit Exceeded" error because CloudKit has a hard limit of 400 items per request.

```
Caught error: <CKError 0x1310f9230: "Limit Exceeded" (27/1020); 
"Your request contains 1360 items which is more than the maximum number of items in a single request (400)">: 
SQLiteData CloudKit Failure
```

This occurs during initial sync or when a large batch of records needs to be synchronized.

### Expected behavior

`SyncEngine` should automatically chunk large batches of records into groups of 400 or fewer items per CloudKit request, similar to how batch operations are handled in the application layer.

### Actual behavior

`SyncEngine` attempts to sync all pending records in a single CloudKit request, which fails when the count exceeds 400 items.

### Reproducing scenario

1. Create a table that accumulates many records 
2. Enable iCloud sync with `SyncEngine` for that table
3. Record hundreds of location points locally
4. Trigger sync

Result: CloudKit sync fails with "Limit Exceeded (27/1020)" error

### Suggested solution

Add automatic chunking in `SyncEngine` to split large batches into 400-item chunks before sending to CloudKit, similar to:

### Destination operating system

iOS 26

### Xcode version information

Xcode 26.2

mbrandonw:
Right now it is possible to run into a `limitExceeded` error when downloading lots of data (such as fresh install) when there are associations that have lots of data. Those associations can only sync when their parent record is available, and if the parent is not available the record ID is cached to be tried later. The problem is that when the record is tried later we load the freshest data from the server via [records(for:)](https://github.com/pointfreeco/sqlite-data/blob/df360b234eec719360b1fa7912efeff54d6eee8e/Sources/SQLiteData/CloudKit/SyncEngine.swift#L1493), and _that_ API must be batched if fetching lots of records.

This PR implements that batching logic and updates the mock cloud database to reject fetching/sending data if there is too much data.

Fixes #333 / #334

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -1523,8 +1523,8 @@
 
       let modifications = (modifications + unsyncedRecords).sorted { lhs, rhs in
         topologicallyAscending(
-          lhsTableName: lhs.recordID.tableName,
-          rhsTableName: rhs.recordID.tableName,
+          lhsTableName: lhs.recordType,
+          rhsTableName: rhs.recordType,
           rootFirst: true
         )
       }
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -1501,10 +1501,10 @@
           }
           var unsyncedRecords: [CKRecord] = []
           for batch in 0...batchCount {
-            let recordIDs = orderedUnsyncedRecordIDs
+            let recordIDsBatch = orderedUnsyncedRecordIDs
               .dropFirst(batch * batchSize)
               .prefix(batchSize)
-            let results = try await syncEngine.database.records(for: Array(recordIDs))
+            let results = try await syncEngine.database.records(for: Array(recordIDsBatch))
             for (recordID, result) in results {
               switch result {
               case .success(let record):
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -24,7 +24,7 @@
     package let tables: [any SynchronizableTable]
     package let privateTables: [any SynchronizableTable]
     let tablesByName: [String: any SynchronizableTable]
-    private let tablesByOrder: [String: Int]
+    package let tablesByOrder: [String: Int]
     let foreignKeysByTableName: [String: [ForeignKey]]
     package let syncEngines = LockIsolated<SyncEngines>(SyncEngines())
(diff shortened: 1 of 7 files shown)


Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -1419,11 +1419,11 @@
     ) async {
       let deletedRecordIDsByRecordType = OrderedDictionary(
         grouping: deletions.sorted { lhs, rhs in
-          guard
-            let lhsIndex = tablesByOrder[lhs.recordType],
-            let rhsIndex = tablesByOrder[rhs.recordType]
-          else { return true }
-          return lhsIndex > rhsIndex
+          topologicallyAscending(
```
mbrandonw: We have a bunch of places we do this sorting logic, and I needed to do it again, so I added a helper for it.

Tests/SQLiteDataTests/CloudKitTests/ForeignKeyConstraintTests.swift
```diff
@@ -869,6 +869,70 @@
...
+
+      /*
+       * Create a parent record in CloudKit database but do not sync to client.
+       * Create many child records in CloudKit database and **do** sync to client.
+       * Sync parent record to client.
+       * => Cached unsaved child records should be batched so as to not run into 'limitExceeded'
+            errors
+       */
+      @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+      @Test func batchAssociations() async throws {
```
mbrandonw: This test failed before the fixes in this PR because it creates 500 reminders and syncs them to the client before the client receives the reminders list. That lead to the limit exceeded error and the system couldn't recover from that.

Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -217,7 +217,7 @@
       tables: [any SynchronizableTable],
       privateTables: [any SynchronizableTable] = []
     ) throws {
-      let allTables = Set((tables + privateTables).map(HashableSynchronizedTable.init))
+      let allTables = OrderedSet((tables + privateTables).map(HashableSynchronizedTable.init))
```
mbrandonw: Unrelated to the core of this PR, but I did want to write a test on our topological sorting of tables, and so needed the order of things to be stable.

stephencelis (approved):

PR #360, opened 2026-01-07
[Proposal] Add a delegate method that handles db conflict for records being inserted from the cloud

hishnash:
While it is nice to assume we can keep our db schema backwards compatible the reality is this leads to issues.

For example we we had a `PostalAddress` table that has an field `enum` for addressType and later wanted to let users extend this the logic step would be to create a new table `AddressType` and run a db migration that creates a row in that table for each enum case and update the  `PostalAddress` to have a column that points to this `AddressType` table with a data migration to set the corresponding FK value into this filed on `PostalAddress`. 

To ensure our local schema is clean we do not want to keep an old `addressType` enum filed alongside a new optional `addressTypeID` and then ensure every reference in the app that lookups up one also checks the other...  so we removed the old `addressType`  field after doing the data migration that created a `AddressType` row for each referenced address type and set the corresponding `addressTypeID` value. 

But now imagine a user that had been using that app on the old version, removed the app from their device and then installed the app again (or more commonly got a new phone and thus got the new version of the app but without any local data).. as they open the app the data migration runs against the local (empty) database.   As they keep the app open the users real data starts to sync in from iCloud, however this data has the old `addressType` enum values attached to it rather than the `UUID` of the new `AddressType` table rows.    The users `PostalAddress` rows will not sync as the insert request will be missing the `addressType` ID that needs to be non optional. 

I propose adding a `handleUpsertFromServerRecord` method to the delegate that is called if there is a `DatabaseError` when trying to `upsert` a local value.  Within this method one can then write any custom logic (such as in this case create a AddressType, if needed, and include that addressTypeID in the insert request).  

---

The other use case this can be important is if the user has 2 devices and they have only updated on one device but make db record changes on the other, the update even that come over the wire may end up not including a value for the new column and thus setting it to null.   This is an issue even if the new column is optional as the row itself might even already have a value set but the update that comes through ends up nuking it.   Being able to run whatever data migration logic we would apply locally to these insert requests (if a DB constraint is triggered) would be a great way to avoid such issues.

---

## Alternatives considered:
 
Have a method that is called before `insert/update/delete` for each DB letting the delegate modify the `CKRecord ` (or even suppress an update or delete).  This is tempting but it would increase the cost of syncing for all users all the time, whereas the implemented solution only kicks in when a db constraint is triggered is much more tightly scoped. 

---
TODO:

- [ ] add examples in doc string of how to extract the needed data from `CKRecord` 
- [ ] add unit test
- [ ] add to one or more of the example projects

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -1908,7 +1908,27 @@
 
           do {
             try $_currentZoneID.withValue(serverRecord.recordID.zoneID) {
-              try #sql(upsert(table, record: serverRecord, columnNames: columnNames)).execute(db)
+                do {
+                    try #sql(
+                      upsert(
+                          table,
+                          record: serverRecord,
+                          columnNames: columnNames
+                      )
+                    ).execute(db)
+                } catch (let error as DatabaseError) {
+                    if let delegate {
+                        try delegate.handleUpsertFromServerRecord(
+                            error: error,
+                            serverRecord: serverRecord,
+                            columnNames: columnNames,
+                            table: T.self,
+                            into: db
+                        )
+                    } else {
+                        throw error
+                    }
+                }
             }
             try UnsyncedRecordID.find(serverRecord.recordID).delete().execute(db)
             try SyncMetadata
--- a/Sources/SQLiteData/CloudKit/SyncEngineDelegate.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngineDelegate.swift
@@ -78,6 +78,35 @@
       _ syncEngine: SyncEngine,
       accountChanged changeType: CKSyncEngine.Event.AccountChange.ChangeType
     ) async
+      
+    
+    /// Handle any DatabaseError thrown while attempting to update or insert a record from iCloud
+    ///
+    /// By default, a sync engine will re-throw this error and end up not syncing the record. To override this behaviour, _e.g._ if your local schema is no longer compatible and you need to apply a custom data migration before inserting, implement this method, and explicitly handle your db updates.
+    ///
+    /// For example, your `PostalAddress` Table may have in the past used an enum field
+    /// for addressType: (residential, commercial etc) but now uses a foreignKey to a
+    /// dedicated table of `AddressType`. However the record being restored from iCloud
+    /// predates this database migration so it attempting to save the string `residential`
+    /// into the `addressType` field rather than the `ID` of the `AddressType` row who's
+    /// name is `residential`, overriding this method and handling update/inserts to `PostalAddress`
+    /// in this case will allow users to restore historic iCloud data when they re-install
+    /// the app even through the schema is no longer backwards compatible:
+    ///   .... EXAMPLE here
+    ///
+    /// Parameters:
+    ///   - error: The DatabaseError that was thrown.
+    ///   - serverRecord: The CKRecord that is being updated or inserted.
+    ///   - columnNames: The column names that contain an update.
+    ///   - table: The local database table.
+    ///   - Database: The database into which the changes should be written.
+    func handleUpsertFromServerRecord<T: Table>(
+        error: DatabaseError,
+        serverRecord: CKRecord,
+        columnNames: some Collection<String>,
+        table: T.Type,
+        into database: Database
+    ) throws
   }
 
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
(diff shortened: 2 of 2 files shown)


mbrandonw:
Hi @hishnash, for this kind of proposal it would help to see a concrete example of how it could be used.

> While it is nice to assume we can keep our db schema backwards compatible the reality is this leads to issues.

Unfortunately the reality is that DB schemas _must_ be backwards compatible, and there really is no way around it. It's the only way to allow for old and new schemas to communicate with each other.

The delegate action you are proposing may work for new schemas interpreting data from old schemas, but it does nothing to help old schemas interpret data from new schemas. And it can't possibly because apps with the old schema were shipped long before we knew the demands on the new schema.

This does mean that over time your schema may become less than ideal because you are forced to make compromised decisions to keep backwards compatibility. Sometimes that can be mitigated with various techniques, but also it is just the reality by having a schema defined on device as well as remotely on iCloud.

> The other use case this can be important is if the user has 2 devices and they have only updated on one device but make db record changes on the other, the update even that come over the wire may end up not including a value for the new column and thus setting it to null. This is an issue even if the new column is optional as the row itself might even already have a value set but the update that comes through ends up nuking it. Being able to run whatever data migration logic we would apply locally to these insert requests (if a DB constraint is triggered) would be a great way to avoid such issues.

I mentioned this in our conversation over Slack, but this is _not_ how SQLiteData is supposed to work, and if you are seeing this behavior it would be a bug. We have test coverage on this exact scenario, and I just [added more](https://github.com/pointfreeco/sqlite-data/pull/361), and so I feel somewhat confident that it works. Each version of the app keeps around a full `CKRecord` of the last known server record, which includes fields that they may not even have represented in their schema. And this is the record that is updated and sent back to iCloud, and so it should never be possible for an old app to update a record and cause data loss.

But if you definitely saw this data loss then we will need a lot more info on how to reproduce it. There could definitely be a bug in our code, and we would rather fix that than consider all new ways to provide compatibility between different schemas.

hishnash:
> You should not make any writes to the database before the sync engine is setup.

So we cant do data migrations (or fixture population) within a traditional DB migration flow?   We would need to create some other follow up hook that runs as a side effect of having just run a schema migration.  Eg if schema migration 0001 just ran then after sync engine starts run schema migration 0001's related data migration.   

 Or would there be a way we can just track what records have been updated/inserted/deleted during the data migration and then tap the sync engine on the shoulder when it starts up (in effect calling the same methods in the sync engine that the db triggers would have called had the changes happened while it was running)?


The reason I mention multiple versions is that when the data migration runs (whenever it is) we cant be certain all changes have been synced up from other devices let alone synced down to the users device that has the new version.  So the user on the old device might well get a prompt telling them to upgrade (this only happens if there is a schema change of course so not ever update) but while that prompt is on screen the that users device might well still have pending changes it has not yet streamed up to iCloud so even through the shame has effectively changed updates (and inserts) with the old schema will still be in flight on the wire.

---
I am a little concerned that as the app evolves if we depend on all migrations to use DB trigger we are going to end up with a very messy schema that is very difficult to reason with.

mbrandonw:
> So we cant do data migrations (or fixture population) within a traditional DB migration flow? We would need to create some other follow up hook that runs as a side effect of having just run a schema migration. Eg if schema migration 0001 just ran then after sync engine starts run schema migration 0001's related data migration.

Any changes made to your database before the `SyncEngine` is running will not be seen by the sync engine, and so will not be sent to iCloud. But also these kinds of "data migrations" are hard to get right. They are going to be run on every device, even if another device already ran them. For example, device 1 could have run the data migration creating record A, B, C, then it deletes A, and when device 2 runs for the first time it will recreate A, B, C, even though A had been deleted by device 1.

This is just the complication of having each device be in control of the schema of the database, and so you have to be careful with these things. You can't just approach the problems exactly as you would in a more traditional database set up prior to synchronization.

> Or would there be a way we can just track what records have been updated/inserted/deleted during the data migration and then tap the sync engine on the shoulder when it starts up (in effect calling the same methods in the sync engine that the db triggers would have called had the changes happened while it was running)?

You can always keep track of the rows you inserted into the DB during the data migration and then _after_ the sync engine is set up do the following:

```swift
let idsCreatedDuringMigration = […]
try await database.write { db in 
  try Record
    .find(idsCreatedDuringMigration)
    .update { $0.id = $0.id }
    .execute(db)
}
```

That will cause the sync engine to see the records change (even though nothing really changed), and will enqueue changes into `CKSyncEngine` to be sent out later. But you still have the potential problem I mentioned above.

> I am a little concerned that as the app evolves if we depend on all migrations to use DB trigger we are going to end up with a very messy schema that is very difficult to reason with.

What do you mean "all migrations to use DB trigger"? You should be designing your schema carefully enough so that these kinds of problems are rare. This should not be happening all of the time.

And again, this is just the complication of having each device be in control of the schema. The same is true of SwiftData, and honestly the problem is much more pernicious over there.

---

And to go back to the original purpose of this PR, you still need to explain exactly how this change will be useful. As far as I can tell, it does not help when older devices receive data from newer schemas. Are you sure this actually solves anything? Can you show a concrete example of how it would be used?

PR #362, opened 2026-01-08
Avoid overwriting local db column values in cases were `CKRecord` has no value.

hishnash:
it appears that if a db record is created (or modified) while the sync engine is not active (such as running a data migration at the same time as the db migration as is common) then when the sync engine starts the `SyncEngine.updateLocalFromSchemaChange` method is called and this in turn courses the `updateQuery` method and here any new columns you may have added are included, however since there is no entry in `CKRecord` for these columns the query is constructed to set this value to `NULL` on conflict.   

This change filters the fields used in the `ON CONFLICT` to only update fields were the `CKRecord` has some values.  I wander if an additional change that might be worth having here is a call that updates the local CKRecord to keep it in sync with the local db so that data migrations are then correctly tracked? 

I have also updated the method that is called when a cloud sync comes in over the wire to ensure that it also does not end up setting columns to NULL within the update block in cases were there is an existing value and the `CKRecord` has not indication of that column existing at all. 

This change assumes that `CKRecord` always persists null against a columnName rather than skipping the column.  I believe this is the correct assumption but maybe for columns with attachments it is not?

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -2046,7 +2046,7 @@
       if columnNamesForUpdate.isEmpty {
           query.append(") ON CONFLICT(\(quote: T.primaryKey.name)) DO NOTHING ")
       } else {
-          query.append(") ON CONFLICT(\(quote: T.primaryKey.name)) DO UPDATE SET ")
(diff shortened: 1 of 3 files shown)


hishnash:
@mbrandonw  I have this branch to be ontop of the latest changes.

PR #363, opened 2026-01-09
Fetch current record before sharing.

mbrandonw:
There's a "quirk" (maybe bug?) in CloudKit where saving a child record will cause iCloud to update the parent's `recordChangeTag` (this part is documented), but that update does not get sent to the `CKSyncEngine` (not documented, possible bug). This can cause us to have a stale `lastKnownServerRecord`, which can cause updates to the parent record to fail. Most of the time that is handled just fine by the conflict resolution mechanism in the library, but when sharing in particular we have to handle the situation more delicately. So now we make sure to fetch the freshest record from iCloud before sharing.

### DIFF
--- a/Tests/SQLiteDataTests/CloudKitTests/SharingTests.swift
+++ b/Tests/SQLiteDataTests/CloudKitTests/SharingTests.swift
@@ -847,6 +847,86 @@
         }
       }
 
+      /*
+       * Create parent record and synchronize.
+       * Create child record and synchronize.
+       * Share parent record.
+       */
+      @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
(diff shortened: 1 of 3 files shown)


Tests/SQLiteDataTests/CloudKitTests/SharingTests.swift
```diff
@@ -847,6 +847,87 @@
         }
       }
 
+      /*
+       * Create parent record and synchronize.
+       * Create child record and synchronize.
+       * Share parent record.
+       * See: https://github.com/pointfreeco/sqlite-data/pull/363
```
stephencelis: Should we start using `@Test(.bug("<url>"))` to track this kind of thing?

stephencelis (approved):

PR #367, opened 2026-01-09
Better support for sharing in previews.

mbrandonw:
We can emulate iCloud sharing in previews by making some updates to our mocks that power the sync engine.


Sources/SQLiteData/CloudKit/Internal/MockCloudDatabase.swift
```diff
@@ -9,6 +9,7 @@
     package let databaseScope: CKDatabase.Scope
     let _container = IsolatedWeakVar<MockCloudContainer>()
     let dataManager = Dependency(\.dataManager)
+    let deletedRecords = LockIsolated<[(CKRecord.ID, CKRecord.RecordType)]>([])
```
mbrandonw: We keep track of records deleted from the mock cloud database so that when fetching changes we can fetch deletions.

stephencelis (approved):

Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
```diff
@@ -25,12 +25,7 @@
       throw InMemoryDatabase()
     }
 
-    let metadatabase: any DatabaseWriter =
-      if url.isInMemory {
-        try DatabaseQueue(path: url.absoluteString)
-      } else {
-        try DatabasePool(path: url.path(percentEncoded: false))
-      }
+    let metadatabase = try DatabasePool(path: url.path(percentEncoded: false))
```
lukaskubanek: @mbrandonw Hi Brandon, I ran into a crash when trying to execute tests on macOS 15. It seems that `URL(string: "file:sqlitedata_icloud?mode=memory&cache=shared").path(percentEncoded: false)` produces an empty string, resulting in `DatabasePool(path: "")`, which crashes trying to activate WAL mode. On macOS 26 (CI) it seems to return `"sqlitedata_icloud"` instead, hiding the issue. Seems like a change in `Foundation.URL` inner guts. I was wondering whether the `DatabaseQueue` branch was intentionally removed or if we should revert this change to continue supporting macOS 15?

PR #369, opened 2026-01-12
Fix metadata observation.

mbrandonw:
Right now some changes to the metadatabase tables will not notify observers because the writes are made with the wrong connection.

### DIFF
--- a/Tests/SQLiteDataTests/CloudKitTests/MetadataTests.swift
+++ b/Tests/SQLiteDataTests/CloudKitTests/MetadataTests.swift
@@ -235,7 +235,7 @@
           │   ),                                                                                    │
           │   share: nil,                                                                           │
           │   _isDeleted: false,                                                                    │
-          │   hasLastKnownServerRecord: true,                                                       │
+          │   _hasLastKnownServerRecord: true,                                                      │
           │   _isShared: false,                                                                     │
           │   userModificationTime: 0                                                               │
           │ )                                                                                       │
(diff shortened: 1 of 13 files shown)


Tests/SQLiteDataTests/CloudKitTests/MetadataTests.swift
```diff
@@ -621,6 +621,72 @@
...
+          database: userDatabase.database
+        )
+        var rows
+        try await $rows.load()
+        #expect(rows == [RecordNameAndIsShared(recordName: "1:remindersLists", isShared: false)])
+
+        _ = try await syncEngine.share(record: remindersList) { _ in }
+
+        try await Task.sleep(for: .seconds(0.5))
+        #expect(rows == [RecordNameAndIsShared(recordName: "1:remindersLists", isShared: true)])
```
mbrandonw: This test fails without these changes.

Tests/SQLiteDataTests/CloudKitTests/MetadataTests.swift
```diff
@@ -621,6 +621,72 @@
...
+        var rows
+        try await $rows.load()
+        #expect(rows == [RecordNameAndIsShared(recordName: "1:remindersLists", isShared: false)])
+
+        _ = try await syncEngine.share(record: remindersList) { _ in }
+
+        try await Task.sleep(for: .seconds(0.5))
+        withKnownIssue("Query observation does not work with generated columns right now") {
+          #expect(rows == [RecordNameAndIsShared(recordName: "1:remindersLists", isShared: true)])
+        }
```
mbrandonw: Observation of generated columns does not currently work.

stephencelis (approved):

PR #380, opened 2026-01-20
Fix 'NULL DEFAULT _' migrations.

mbrandonw:
This is an alternative fix to #362 

When performing a migration of the form "ADD COLUMN … DEFAULT _" (i.e. a nullable column), then we want the default to be inserted into the local database and stay there after the sync engine has fully started up. Currently there is a bug where that data gets NULL'd out after start up, and this PR fixes that.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -2004,8 +2004,7 @@
       let nonPrimaryKeyChangedColumns =
         changedColumnNames
         .filter {
-          $0 != T.primaryKey.name
-          && record.hasSet(key: $0)
+          $0 != T.primaryKey.name && record.hasSet(key: $0)
         }
       guard
         !nonPrimaryKeyChangedColumns.isEmpty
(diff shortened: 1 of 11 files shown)


mbrandonw:
@hishnash Can you check if this fixes the problem you were fixing in #362 ?

hishnash:
what about the UPSERT operation that is called when another device makes a change.

What would happen if a device on an older version makes a change and it is synced to a device with the new schema.

What column names are passed to `private func upsert<T>`   would it include the column name for the column that the other device does not have any value for and thus result in overwriting null or would it skip that column in the update portion of the `on conflict`?

Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -2436,6 +2439,13 @@
...
   ) -> QueryFragment {
     let allColumnNames = T.TableColumns.writableColumns.map(\.name)
+      .filter {
+        record.hasSet(key: $0)
+      }
+    guard !allColumnNames.isEmpty
+    else {
+      return ""
+    }
     let hasNonPrimaryKeyColumns = columnNames.contains { $0 != T.primaryKey.name }
```
hishnash: is it possible that `columnNames` here somehow ends up with the column name that is not in the filtered `allColumnNames` and thus would result in a SQL error as `columnNames` is used in the  `ON CONFLICT DO SET` option.   Maybe `columnNames` need to drop any names that are not found in `allColumnNames` (I know there is some other method doing this further up before upsert<T> is called but that method does not consider the `hasSet(key: $0)` )

stephencelis (approved):

PR #385, opened 2026-01-23
Expected failure for permission error deleting root shared record.

mbrandonw:
Something came up in Slack that shows we are enforcing permission settings even when deleting the root shared record. Ideally we would allow deleting the root record because that would just translate to removing yourself from the share, but instead we get an error when the delete cascades to child records.

I don't have a fix yet, but wanted to get a test failure in place (wrapped in `withKnownIssue`) so that we can investigate later.


Rspoon3:
Additional slack [thread](https://pointfreecommunity.slack.com/archives/C08CZUG4Z9D/p1771435029183079).

PR #387, opened 2026-01-24
Fixes to date synchronization.

mbrandonw:
Right now it's possible for dates to change by round tripping them to the database and back into Swift. This is due to the fact that `Date` is the de facto standard for dates in Swift, and `Date` represents time with a double.

This is a test from the PR that failed prior to the changes to show the problem:

```swift
  @Test func roundtrip() throws {
    let date = Date(timeIntervalSinceReferenceDate: 793109282.061)
    let insertedRecord = try database.write { db in
      try Record.insert { Record.Draft(date: date) }
      .returning(\.self)
      .fetchOne(db)!
    }
    let updatedRecord = try database.write { db in
      try Record
        .update(insertedRecord)
      .returning(\.self)
      .fetchOne(db)!
    }
    #expect(insertedRecord.date == date)
    #expect(insertedRecord.date == updatedRecord.date)
  }
```

To fix we need to `nextUp` the date before turning it into an ISO8601 formatted string.

### DIFF
--- a/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
+++ b/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
@@ -146,6 +146,10 @@
 
   @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
   extension CKRecord {
+    func hasSet(key: String) -> Bool {
+      self.encryptedValues["\(CKRecord.userModificationTimeKey)_\(key)"] != nil
+    }
+
     @discardableResult
     package func setValue(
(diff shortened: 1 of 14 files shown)


stephencelis (approved):

mbrandonw:
Currently our stance is that we should let each of SQLite and iCloud deal with the types that makes the most sense for them. So that means ISO8601 strings for dates in SQLite, and `Date` (i.e. `Double`) for iCloud. That makes it so that theoretically someone can query their iCloud database using regular CloudKit tools if they wanted, for whatever reason. T

PR #392, opened 2026-02-06
Fix mock CKAsset temp path in previews

Rspoon3:
## Summary
- Write CKAsset files in MockCloudDatabase using the data manager temporary directory instead of root.
- Add a test to assert assets are stored under the temporary directory.

## Why
Preview/test CloudKit sharing with BLOB columns fails because assets are written to /<uuid> (read-only).

## Error
"🟣 Scorekeeper/GameFeature.swift:90: Caught error: Error Domain=NSCocoaErrorDomain Code=642 \"You can't save the file \"e8525bd6-4718-4db9-827b-67f72459f877\" because the volume \"Macintosh HD - Data\" is read only.\" UserInfo={NSFilePath=/e8525bd6-4718-4db9-827b-67f72459f877, NSURL=e8525bd6-4718-4db9-827b-67f72459f877 -- file:///, NSUnderlyingError=0x60000111bcc0 {Error Domain=NSPOSIXErrorDomain Code=30 \"Read-only file system\"}}"

## Steps to Reproduce
1. In `Scorekeeper/Schema.swift`, add a `Data` column to `Game`:
   ```swift
   @Table struct Game: Identifiable {
     let id: UUID
     var title = ""
     var attachment: Data = Data()
   }
   ```
2. Update migration: `"attachment" BLOB NOT NULL DEFAULT x''`
3. Seed with a non-empty value: `Game.Draft(id: UUID(1), title: "Family gin rummy", attachment: Data([0x01]))`
4. Build/run preview and trigger sharing (`syncEngine.share(record:)`).

## Expected Result
Sharing succeeds (or the mock sharing view appears in preview) without filesystem errors.

## Actual Result
Sharing fails during `syncEngine.share(record:)` with NSCocoaErrorDomain 642, and the file path is rooted at `/`.

## Notes
- Does not occur before adding the `Data` column.
- CKAsset staging uses a temp directory that resolves to `/` in preview/test contexts.
- Suggests preview/test should use a writable temp directory (e.g. `NSTemporaryDirectory`) or allow a dependency override for asset temp directory.

## Testing
- Added MockCloudDatabaseTests.assetsUseTemporaryDirectory


mbrandonw (approved):
Hey @Rspoon3, thanks for this! We made a few changes (using `InMemoryDataManager` in previews and updated the test), but will merge and release soon.

PR #409, opened 2026-03-02
Fix database-access for sharing records 

baronfac:
**Issue**

When opening `CloudSharingView` for a record that was not created by the current user, the following error occurs:
```
PrivateDB can't be used to access another user's zone
```
This is reproducible in the `RemindersApp` demo.
The issue is in CloudKitSharing.share(record:configure:), which always accesses privateCloudDatabase (line 178):
```
let (saveResults, _) = try await container.privateCloudDatabase.modifyRecords(
  saving: [sharedRecord, lastKnownServerRecord],
  deleting: []
)
```
This fails for records owned by other users, since those exist in the shared database, not the private one.
 
**Fix:** 
Resolve the database from the record's ID instead:
```
let (saveResults, _) = try await container.database(for: sharedRecord.recordID).modifyRecords(
  saving: [sharedRecord, lastKnownServerRecord],
  deleting: []
)
```

This ensures the correct database scope is used regardless of record ownership, and CloudSharingView opens as expected.


mbrandonw (approved):
Hi @baronfac, great fix and sorry for the delayed response!

PR #410, opened 2026-03-03
Fix metadatabase crash in tests on macOS 15

lukaskubanek:
This PR reverts a change made in #367, which dropped the `DatabaseQueue` path for in-memory database creation for the metadatabase, leading to a crash when running the test suite on macOS 15. See [this comment](https://github.com/pointfreeco/sqlite-data/pull/367#discussion_r2867236715) for more details.

The fix can be verified by running the test suite on macOS 15 (no longer configured in the CI).

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -25,7 +25,12 @@
       throw InMemoryDatabase()
     }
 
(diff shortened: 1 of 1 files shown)


lukaskubanek:
While revisiting this, I found the issue is not limited to macOS 15. On macOS 26 and later, the metadatabase silently ends up on disk instead of in memory. Closing in favor of #508, which reframes the problem accordingly and adds a regression test.

PR #411, opened 2026-03-03
Set `CKRecord.modificationDate` in `MockCloudDatabase`

lukaskubanek:
## Motivation

SQLiteData’s tests leverage `MockCloudDatabase`, which currently doesn’t populate the `modificationDate` on saved records. Since this value is checked by the sync machinery, the tests effectively exercise different behavior than in production.

This has been first noticed while [documenting the current sync behavior](https://github.com/structuredpath/sqlite-data-sync-notes/blob/main/CurrentBehavior.md) while elaborating on custom conflict resolution support (search for _modificationDate_) and further confirmed as a known issue [here](https://github.com/pointfreeco/sqlite-data/pull/386#issuecomment-3813576376).

This is another step toward bringing `MockCloudDatabase` closer to CloudKit’s actual behavior, and an intermediate step toward adding support for configurable conflict resolution.

## Implementation

Similar to record change tags, `MockCloudDatabase` now assigns incrementing dates to records saved in `modifyRecords(…)`, mirroring CloudKit’s server-side behavior. However, since the codebase checks the actual `CKRecord.modificationDate`, this PR, unlike the current handling of change tags, swizzles the Objective-C getter to return the value stored in `CKRecord._modificationDate` via associated object storage. This is safe in production, as the global swizzle is only triggered when setting `CKRecord._modificationDate`, which is limited to `MockCloudDatabase`, used exclusively in tests. See the next section for alternative approaches that were considered.

Similar to the change tag, the injected modification date must survive the SQLite roundtrips, which is why both `CKRecord.SystemFieldsRepresentation` and `CKRecord._AllFieldsRepresentation` were extended to persist `_modificationDate` alongside the existing `_recordChangeTag` using `NSKeyedArchiver`/`NSKeyedUnarchiver`.

My assumption is that when saving a record, CloudKit only sets the `modificationDate` on that record and it doesn’t “touch” the parent to modify its `modificationDate`, as it apparently does with the `recordChangeTag`. I’m not 100% sure about this behavior, though.

## Alternative Approaches

One way to circumvent swizzling would be to subclass `CKRecord`, override `modificationDate`, and return a mocked record from `MockCloudDatabase.modifyRecords(…)`. Not only is it tricky to change the class of the copied record due to its internal implementation, but CloudKit now logs the following message when attempting to subclass `CKRecord`:

```
BUG IN CLIENT OF CLOUDKIT: CKRecord should not be subclassed (`…MockRecord`). If it is, Sendable may no longer apply. This will become an error in a near future release.
```

Another idea would be to use a facade property that returns the mock version and falls back to the real value if not set. Something like `CKRecord.effectiveModificationDate`. However, this would pollute the API and require all call sites in the library to go through this property, which is quite error-prone.

I’ve come to the conclusion that swizzling is the right tool for the job here, but I’m open to discussing further.

Also, I initially started with using `@Dependency(\.date.now)` for providing dates set to the `modificationDate` in `MockCloudDatabase`. However, giving the mock database control over the values felt more natural and better aligned with how record change tags are handled, so I ended up using the same pattern.

## Impact on Test

When running the test suite with this change, there is one test failing, `MergeConflictTests.clientRecordUpdatedBeforeServerRecord()`:

```diff
@@ −10,9 +10,9 @@
         dueDate🗓️: 0,
         id: 1,
         id🗓️: 0,
         isCompleted: 1,
-        isCompleted🗓️: 30,
+        isCompleted🗓️: 60,
         priority🗓️: 0,
         remindersListID: 1,
         remindersListID🗓️: 0,
         title: "Buy milk",
```

A single timestamp diff in the last assertion might seem minor, but there is an important conceptual decision to be made. This flow might not be immediately obvious, so let me break it down:

| Step | Event | `SyncMetadata.userModificationTime` | `CKRecord.🗓️` | `CKRecord.title🗓️` | `CKRecord.isCompleted🗓️` |
|---|---|---|---|---|---|
| 1 | Initial setup | 0 | 0 | 0 | 0 |
| 2 | Simulate server-side change of title="Buy milk" | n/a | 60 | 60 | 0 |
| 3 | Update locally isCompleted=true | 30 | 60 | 60 | 0 |
| 4 | Process local changes & hit `.serverRecordChanged` error | **60** | 60 | 60 | 0 |
| 5 | Process local changes | 60 | 60 | 60 | **60** |
| 6 | Fetch record zone changes | 60 | 60 | 60 | 60 |

> [!NOTE]
> The `CKRecord` timestamps in the table refer to the server state, i.e. the `MockCloudDatabase`, not the last-known server record stored in the `SyncMetadata`.

In step 4, when hitting the `.serverRecordChanged` error, the server record’s `CKRecord.userModificationTime` is promoted to be the new `SyncMetadata.userModificationTime` following this logic: `max(SyncMetadata.userModificationTime, CKRecord.userModificationTime)`, `max(30, 60)`, `60`.

This means that in step 5, when processing the local changes again after resolving the conflict, the `isCompleted` change is now considered to happen at `t=60`. This means that the original timestamp of the change is lost and I’m wondering if this behavior, which already occurs in production (!), is desired.

I’d say yes, because with support for configurable conflict resolution via a three-way merge, an entirely new value may emerge rather than selecting either the local or server value. And in that case it should be marked as changed at the current `SyncMetadata.userModificationTime`.

Do you agree? Can we accept the change in the test?

## Next Steps

1. Refactor the merge conflict tests with a broader set of scenarios. → Already prepared as a draft in [#414](https://github.com/pointfreeco/sqlite-data/pull/414), based on this PR.
2. Mock `CKRecord.recordChangeTag` via the same swizzling pattern, including aligning the mock change tag format with what CloudKit actually uses (base-36 strings). → Ready to submit once this PR is merged.


lukaskubanek:
Just a heads-up: I have [a follow-up ready](https://github.com/structuredpath/sqlite-data/commit/a11ce8a5fabe64f141894c573bb3a84ce866fb31) that extends the swizzling infrastructure introduced in this PR to also mock `CKRecord.recordChangeTag`, using base-36 change tags to match CloudKit's format. Happy to submit it as a separate PR once this one is merged.

lukaskubanek:
I’ve updated the PR with the latest changes from `main`.

Just a reminder, this is another step towards improving the `MockCloudDatabase` and aligning it with actual CloudKit behavior after #507 and #510.

PR #420, opened 2026-03-11
Fix SyncEngine retain cycle causing file descriptor exhaustion

seanmrich:
Follow up to earlier [discussion](https://github.com/pointfreeco/sqlite-data/discussions/416).

SyncEngine's `@DatabaseFunction` instance methods (`didUpdate`, `didDelete`) capture `self` strongly via their projected values. Since `db.add(function:)` retains these via `Unmanaged.passRetained`, this creates a cycle:

    SyncEngine → userDatabase → database functions → self (SyncEngine)

This prevents deallocation, leaking file descriptors until the process hits "SQLite error 14: unable to open database file". This can be an issue if running a a large number of unit tests that bootstrap a sync engine.

The fix introduces `FunctionContext`, a lightweight proxy that holds a weak reference (via `IsolatedWeakVar`) back to `SyncEngine`. Only `didUpdate` and `didDelete` are moved to `FunctionContext` — these are the two instance methods that capture `self`. Module-level functions (`currentZoneName`, `currentOwnerName`, `currentTime`, `hasPermission`) and the static `isSynchronizing` don't participate in the cycle and are unchanged.

When `SyncEngine` is deallocated the weak reference nils out and the database functions safely no-op via `guard let syncEngine else { return }`.

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Triggers.swift
@@ -236,24 +236,25 @@
 
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
   extension SyncMetadata {
-    static func callbackTriggers(for syncEngine: SyncEngine) -> [TemporaryTrigger<Self>] {
+    static func callbackTriggers(for context: FunctionContext) -> [TemporaryTrigger<Self>] {
       [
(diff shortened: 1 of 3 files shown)


stephencelis:
@seanmrich Are you interested in trying to tackle that? If not no worries. It's on our list of things to do, which is ever growing :smile:

PR #421, opened 2026-03-13
Fix CloudKit data loss on delete-then-reinsert

### LINKED ISSUE #418: CloudKit data loss after deleting then re-inserting record with the same UUID
### Description

I stumbled across this issue after refactoring an ItemAsset-like table to no longer have its own UUID and instead use the foreign key into its parent table (Item.ID), as suggested in the documentation's [uniqueness constraints section](https://swiftpackageindex.com/pointfreeco/sqlite-data/main/documentation/sqlitedata/cloudkit#Uniqueness-constraints).

Pre-refactor (distinct ItemAsset.ID UUID field), the process to update an ItemAsset record consisted of two-steps: 
1. Delete all ItemAsset records having a specific itemID foreign key (there could be multiple records and the goal was to have just one) 
2. Insert a fresh ItemAsset record.

Post-refactor (combined PK/FK ItemAsset.itemID), the initial step to delete should no longer be necessary (only one ItemAsset record per Item record, enforced by new PK), however, I missed this, and the delete step remained in place. So the steps remained roughly the same:
1. Delete ItemAsset record having a specific itemID PK+FK
2. Upsert ItemAsset record.

With the new delete-then-reinsert setup, when an ItemAsset record exists and is updated, **only the deletion of the record is synchronized to CloudKit, not the insert, resulting in unexpected data loss**. The ItemAsset record remains visible on the device which performed the update, but will be deleted on all other devices after the next CloudKit sync.

I've reproduced this behavior with a simple addition to the CloudKitDemo example project (introducing a CounterAsset table and "Update asset" button):
1. Draft PR with 100% repro setup: https://github.com/pointfreeco/sqlite-data/pull/417
2. [Repro video](https://github.com/user-attachments/assets/21663401-7104-4913-8d77-b2d9e593e0b8)
   * Observe alternating (data loss) -> (asset present) -> (data loss)... pattern. Presumably every other "Update asset" succeeds because there's no CloudKit record to delete anymore, so the insert makes its way through.

Note: another precondition of this behavior is that delete and reinsert events need to occur within a couple seconds of each other (to be part of the same sync batch). If they are spaced out further than that and end up in separate sync batches, the issue does not reproduce.

In my case, this behavior was easy to work around (just remove the unnecessary delete), but I think this still warrants either a fix or at least an update to documentation and Point-Free Way skills warning against this pattern.

### Checklist

- [x] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [x] I have determined whether this bug is also reproducible in a vanilla GRDB project.
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/sqlite-data/issues) or [discussion](https://github.com/pointfreeco/sqlite-data/discussions).

### Expected behavior

Ideally, the local delete-then-reinsert events in this scenario should be collapsed to a single CloudKit **update** event (not delete). But if that's infeasible, or this really does just represent an anti-pattern, then there should be a developer-facing error/issue thrown when SyncEngine detects this situation.

### Actual behavior

Copy/pasted from Description:

With the new delete-then-reinsert setup, when an ItemAsset record exists and is updated, **only the deletion of the record is synchronized to CloudKit, not the insert, resulting in unexpected data loss**. The ItemAsset record remains visible on the device which performed the update, but will be deleted on all other devices after the next CloudKit sync.

I've reproduced this behavior with a simple addition to the CloudKitDemo example project (introducing a CounterAsset table and "Update asset" button):
1. Draft PR with 100% repro setup: https://github.com/pointfreeco/sqlite-data/pull/417
2. [Repro video](https://github.com/user-attachments/assets/21663401-7104-4913-8d77-b2d9e593e0b8)
   * Observe alternating (data loss) -> (asset present) -> (data loss)... pattern. Presumably every other "Update asset" succeeds because there's no CloudKit record to delete anymore, so the insert makes its way through.

Note: another precondition of this behavior is that delete and reinsert events need to occur within a couple seconds of each other (to be part of the same sync batch). If they are spaced out further than that and end up in separate sync batches, the issue does not reproduce.

### Reproducing project

Draft PR: https://github.com/pointfreeco/sqlite-data/pull/417

### SQLiteData version information

65502acdb033ec8025a0bcc443abf2f4ca0598f9

### Sharing version information

2.7.4

### GRDB version information

7.9.0

### Destination operating system

iOS 26

### Xcode version information

Xcode 26.3

### Swift Compiler version information

```shell

```

jsutula:
Fixes #418 

### Root cause
Reinsertion of a record after a delete does not produce a new pending `.saveRecord` change entry due to missing undelete/reinsert state detection in `PrimaryKeyedTable.afterInsert` trigger. No new `.saveRecord` means that only the `.deleteRecord` is propagated.

### Solution

To fulfill the expected behavior of reinsertion resulting in a `.saveRecord` and assigning fresh timestamps to all fields, several changes are needed: one prerequisite change to test infrastructure `MockSyncEngineState`, and a four-part change to detect and handle reinsertion properly.

#### **Prerequisite change necessary for writing accurate tests**: Update `MockSyncEngineState` to deduplicate changes across types like the real `CKSyncEngine.State`

From `CKSyncEngine.State.add(pendingRecordZoneChanges:)` [documentation](https://developer.apple.com/documentation/cloudkit/cksyncengine-5sie5/state-swift.class/add(pendingrecordzonechanges:)?changes=_2,_2):
> Note
> The order in which you apply record zone changes is important.
> For example:
> * If you add `.saveRecord(recordA)` then `.deleteRecord(recordA)`, the sync engine 
> discards the save and sends only the delete change.
> * If you add `.deleteRecord(recordA)` then `.saveRecord(recordA)`, the sync engine
> discards the delete and sends only the save change.

`MockSyncEngineState.add(pendingRecordZoneChanges:)` deduplicates fully identical changes (type+ID) by virtue of the underlying `OrderedSet`, but does not deduplicate when the type (save vs delete) is different, as is described above with `CKSyncEngine.State`. `MockSyncEngineState.add(pendingDatabaseChanges:)` has a similar issue. The fix: update both to deduplicate based on ID alone which will allow for the cross-type deduplication described above, and add tests.

#### 1. Replace `SyncMetadata._isDeleted` with enum value `SyncMetadata._pendingStatus`
The `_isDeleted` field does not provide enough information about the current pending/transition status of the record. Introduce new enum value field `_pendingStatus` to replace `_isDeleted` with two cases: `.deleted` (soft-deleted) and `.reinserted` (row re-added while soft-deleted). 

#### 2. Detect reinsertion of soft-deleted record, update `_pendingStatus` and queue new `.saveRecord` change
Modify `PrimaryKeyedTable.afterInsert` trigger to update the `SyncMetadata` record when reinsertion of a soft-deleted row is detected (insert comes through while `_pendingStatus` is `.deleted`). Updates include:
* Setting `_pendingStatus` to `.reinserted`
   * This status will be used to trigger full re-hydration of `_lastKnownServerRecordAllFields` to be based on local, reinserted state when preparing an outbound record (see `#3` below) and before processing inbound server record (see `#4` below).
* Updating `userModificationTime`
   * Will be used as the modified time for all fields in the reinserted row when hydrating `_lastKnownServerRecordAllFields`

Hook up new `SyncMetadata.afterReinsertTrigger` to listen for reinsertions. This trigger is responsible for queuing a `.saveRecord` change via `syncEngine.$didUpdate` (the same as what happens after a normal insert or update).

#### 3. In `nextRecordZoneChangeBatch` (outbound record preparation), ensure record is built with reinserted row values

When preparing an outbound record for a `.reinserted` row, use `SyncMetadata.lastKnownServerRecord` (system fields only) as the base for building the record. Using `lastKnownServerRecord` as the base (vs `_lastKnownServerRecordAllFields` which would contain stale user field data) ensures the resulting all fields record is built up entirely from the current contents of the reinserted row + reinsert `userModificationTime`.

Additionally, as a cleanup step, upon saving the newly built record back into the metadata row, reset metadata `_pendingStatus` to nil. This ensures this "rebuild record from reinserted row" behavior will occur only once per reinsertion.

#### 4. In `upsertFromServerRecord` (inbound record processing), ensure local `allFields` state reflects reinserted row values prior to server record application

Since `upsertFromServerRecord` can occur *before* `nextRecordZoneChangeBatch`, we also need to ensure that the local metadata 'allFields' state is rebuilt and accurate when we encounter any `.reinserted` rows in this code path.

So before reconciling local metadata and user row with server record data, check if we're dealing with a `.reinserted` row, and if so, update the local `allFields` record to reflect all reinserted row values + reinsert `userModificationTime`. And similarly to the outbound path, clean up by setting `_pendingStatus` to nil at the conclusion of the upsert.

### Testing
In addition to new unit tests, performed a manual test with https://github.com/pointfreeco/sqlite-data/pull/417/changes/9ea5bdd88b79d7c9db88021e568c8a66789a21fa applied to verify that the data loss issue no longer occurs. Result (compare with [before](https://github.com/user-attachments/assets/173f1010-9353-43e6-a2ee-b0455800e85f)):

https://github.com/user-attachments/assets/da7d2907-dbda-4e21-be6d-d1512299bad6

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -125,6 +125,30 @@
       )
       .execute(db)
     }
+    migrator.registerMigration("Replace _isDeleted with _pendingStatus") { db in
+      try #sql(
+        """
+        ALTER TABLE "\(raw: .sqliteDataCloudKitSchemaName)_metadata"
+        ADD COLUMN "_pendingStatus" INTEGER
+        """
+      )
+      .execute(db)
+      try #sql(
+        """
+        UPDATE "\(raw: .sqliteDataCloudKitSchemaName)_metadata"
+        SET "_pendingStatus" = 0
+        WHERE "_isDeleted" = 1
+        """
+      )
+      .execute(db)
+      try #sql(
+        """
+        ALTER TABLE "\(raw: .sqliteDataCloudKitSchemaName)_metadata"
+        DROP COLUMN "_isDeleted"
+        """
+      )
+      .execute(db)
+    }
     #if DEBUG
       try metadatabase.read { db in
         let hasSchemaChanges = try migrator.hasSchemaChanges(db)
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -141,13 +141,6 @@
         """
       )
       .execute(db)
-      try #sql(
-        """
-        ALTER TABLE "\(raw: .sqliteDataCloudKitSchemaName)_metadata"
-        DROP COLUMN "_isDeleted"
-        """
-      )
-      .execute(db)
     }
     #if DEBUG
       try metadatabase.read { db in
--- a/Sources/SQLiteData/CloudKit/Internal/MockSyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/MockSyncEngine.swift
@@ -136,12 +136,12 @@
   package final class MockSyncEngineState: CKSyncEngineStateProtocol {
     package let changeTag = LockIsolated(0)
     package let _pendingRecordZoneChanges = LockIsolated<
-      OrderedSet<CKSyncEngine.PendingRecordZoneChange>
-    >([]
+      OrderedDictionary<CKRecord.ID, CKSyncEngine.PendingRecordZoneChange>
+    >([:]
     )
     package let _pendingDatabaseChanges = LockIsolated<
-      OrderedSet<CKSyncEngine.PendingDatabaseChange>
-    >([])
+      OrderedDictionary<CKRecordZone.ID, CKSyncEngine.PendingDatabaseChange>
+    >([:])
     private let fileID: StaticString
     private let filePath: StaticString
     private let line: UInt
@@ -160,11 +160,11 @@
     }
 
     package var pendingRecordZoneChanges: [CKSyncEngine.PendingRecordZoneChange] {
-      _pendingRecordZoneChanges.withValue { Array($0) }
+      _pendingRecordZoneChanges.withValue { Array($0.values) }
     }
 
     package var pendingDatabaseChanges: [CKSyncEngine.PendingDatabaseChange] {
-      _pendingDatabaseChanges.withValue { Array($0) }
+      _pendingDatabaseChanges.withValue { Array($0.values) }
     }
 
     package func removePendingChanges() {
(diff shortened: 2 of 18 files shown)


aehlke:
This sounds like a pretty critical issue - easy to trigger and results in data loss

nagra:
Any updates on this? I'm pinned against this commit at the moment 😬

jsutula:
I haven't received feedback on this approach from Brandon or Stephen, but I know it's on their radar. This branch is overdue for a rebase which I'll do now.

@nagra Since this change modifies `SyncMetadata` schema (adds `_pendingStatus`, removes `_isDeleted`), be very cautious pinning against this for anything more than local development.

For example, if you’ve already pinned to this and distributed widely via TestFlight / App Store, but the final version of this fix ends up not needing the schema change, then in order to adopt the official fix you would have to spin up a fork + custom patch to ensure that `_isDeleted` is re-added if it was previously removed. And you’d potentially have to maintain this fork for a while - until you’re sure that all existing app installations have upgraded.

Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
```diff
@@ -141,13 +141,6 @@
         """
       )
       .execute(db)
-      try #sql(
-        """
-        ALTER TABLE "\(raw: .sqliteDataCloudKitSchemaName)_metadata"
-        DROP COLUMN "_isDeleted"
-        """
-      )
-      .execute(db)
```
jsutula, line 150: @nagra - after my rebase, I pushed this additional commit so that the old `_isDeleted` column is no longer dropped. This would make it marginally safer to pin to this commit and deploy, but I still wouldn't recommend it since the risk of schema divergence still remains, and it'd be a pain to clean that up in the future.

nagra:
Been another 3 weeks, is this no longer an issue for most people? 🤔

PR #425, opened 2026-03-20
Rename CKRecord.setValue overload to setBytes to fix unexpected Xcode 26.4 overload resolution

jsutula:
## Problem
Unit tests are failing with Xcode 26.4 (I'm on Xcode 26.4 RC 17E192).

When calling `CKRecord.setValue(_ newValue:forKey:at)` with a value of bytes (`[UInt8]`), building with Xcode 26.4 results in an unexpected `setValue` overload being invoked, causing test failures in `SchemaChangeTests` and `AssetsTests`.

**With Xcode 26.3**, the expected overload accepting `[UInt8]` is invoked. This overload saves the bytes in a `CKAsset` and then saves the asset as a regular field on the record.
https://github.com/pointfreeco/sqlite-data/blob/65502acdb033ec8025a0bcc443abf2f4ca0598f9/Sources/SQLiteData/CloudKit/CloudKit%2BStructuredQueries.swift#L193-L194

**With Xcode 26.4**, the overload accepting `some CKRecordValueProtocol & Equatable` is invoked instead. This overload writes the value directly to `encryptedValues`, no `CKAsset`.
https://github.com/pointfreeco/sqlite-data/blob/65502acdb033ec8025a0bcc443abf2f4ca0598f9/Sources/SQLiteData/CloudKit/CloudKit%2BStructuredQueries.swift#L154-L155

## Solution

Rename the overload which accepts bytes to `setBytes` and update all callers. The new name sidesteps the compiler having to resolve the overload and better reflects there being a distinct behavioral difference compared to plain `setValue`.

## Testing
With Xcode 26.4 selected, `swift test`. All tests pass including previously failing `SchemaChangeTests` and `AssetsTests`.

### DIFF
--- a/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
+++ b/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
@@ -190,7 +190,7 @@
     }
 
     @discardableResult
-    package func setValue(
+    package func setBytes(
       _ newValue: [UInt8],
       forKey key: CKRecord.FieldKey,
       at userModificationTime: Int64
@@ -252,7 +252,7 @@
           let value = Value(queryOutput: row[keyPath: keyPath])
           switch value.queryBinding {
           case .blob(let value):
-            setValue(value, forKey: column.name, at: userModificationTime)
+            setBytes(value, forKey: column.name, at: userModificationTime)
           case .bool(let value):
             setValue(value, forKey: column.name, at: userModificationTime)
           case .double(let value):
--- a/Tests/SQLiteDataTests/CloudKitTests/AssetsTests.swift
+++ b/Tests/SQLiteDataTests/CloudKitTests/AssetsTests.swift
@@ -285,7 +285,7 @@
           recordID: RemindersListAsset.recordID(for: 1)
         )
         remindersListAssetRecord.setValue("1", forKey: "id", at: now)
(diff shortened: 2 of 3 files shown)


Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
```diff
@@ -252,7 +252,7 @@
           let value = Value(queryOutput: row[keyPath: keyPath])
           switch value.queryBinding {
           case .blob(let value):
-            setValue(value, forKey: column.name, at: userModificationTime)
+            setBytes(value, forKey: column.name, at: userModificationTime)
```
jsutula: @stephencelis I was hoping the impact of this bug would be in tests only, but I had to change this line here in `CKRecord.update` to fix things, and unfortunately it's called by `SyncEngine`. I believe this means that anyone who's already been building/running/deploying with Xcode 26.4 will have to contend with this behavioral change having impacted how data is stored (blobs directly in the record vs `CKAsset`), and address any side effects stemming from that. Can you confirm if this is something we need to be concerned about?

I also see that you made the change to support Xcode 26.4 last month with release 1.6.0. Were the tests passing with early Xcode 26.4 betas? I jumped straight from 26.3 to 26.4 RC, so it's possible the behavior change may have been introduced in a later beta build or RC.

PR #426, opened 2026-03-20
Add metadatabase URL override for CloudKit sync

byMohamedali:
## Summary

Adds an optional `metadatabaseURL` override to SQLiteData's CloudKit APIs so callers can choose where the CloudKit metadata database lives.

## Changes

- Add `metadatabaseURL: URL? = nil` to:
  - `SyncEngine.init(for:tables:privateTables:containerIdentifier:metadatabaseURL:defaultZone:startImmediately:delegate:logger:)`
  - `Database.attachMetadatabase(containerIdentifier:metadatabaseURL:)`
- Preserve existing behavior when no override is provided.
- Create the parent directory of the resolved metadata URL instead of always creating `Application Support`.

## Motivation

Some platforms, notably tvOS, may not allow the current hardcoded `Application Support` behavior for this metadata database. The main user database may already be configured to live somewhere else, but CloudKit setup still fails because the metadata store is not configurable.

This change keeps the default behavior intact for existing apps while allowing callers to explicitly provide a writable location when needed.

## Example

```swift
let metadataURL = FileManager.default
  .urls(for: .cachesDirectory, in: .userDomainMask)
  .first?
  .appendingPathComponent(".SQLiteData.metadata.sqlite")

var configuration = Configuration()
configuration.prepareDatabase { db in
  try db.attachMetadatabase(metadatabaseURL: metadataURL)
}

let database = try SQLiteData.defaultDatabase(
  path: databaseURL.path,
  configuration: configuration
)

let syncEngine = try SyncEngine(
  for: database,
  privateTables: MyTable.self,
  metadatabaseURL: metadataURL
)
```

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -13,11 +13,12 @@
       open "\(url.path(percentEncoded: false))"
       """
     )
-    try FileManager.default.createDirectory(
-      at: .applicationSupportDirectory,
-      withIntermediateDirectories: true
-    )
-
+    #if !os(tvOS)
+      try FileManager.default.createDirectory(
+        at: .applicationSupportDirectory,
+        withIntermediateDirectories: true
+      )
+    #endif
     @Dependency(\.context) var context
     guard !url.isInMemory || context != .live
     else {
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -2239,10 +2239,12 @@
         containerIdentifier: containerIdentifier
       )
       let path = url.isInMemory ? url.absoluteString : url.path(percentEncoded: false)
-      try FileManager.default.createDirectory(
-        at: .applicationSupportDirectory,
-        withIntermediateDirectories: true
-      )
+      #if !os(tvOS)
+        try FileManager.default.createDirectory(
+          at: .applicationSupportDirectory,
+          withIntermediateDirectories: true
+        )
+      #endif
       let database: any DatabaseWriter =
         url.isInMemory
         ? try DatabaseQueue(path: path)


byMohamedali:
Yes sorry for late answer, on tvOS we can use only this 
`FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first` path and the app crash when trying to create the SyncEngine sql file, have you tried on tvOS ?

Armenm:
I was having the same issue on tvOS.

Metadatabase connection:
open "/var/mobile/Containers/Data/Application/EA2A5822-43D4-4D8D-A18A-B51577A8284C/Library/Caches/AppName/.playlist-credentials.metadata-iCloud.***.sqlite"
[CloudKit] sync unavailable
- Container: iCloud.***
- Domain: NSCocoaErrorDomain
- Code: 513
- Description: You don’t have permission to save the file “Application Support” in the folder “Library”.
- UserInfo: ["NSUnderlyingError": Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted", "NSURL": file:///var/mobile/Containers/Data/Application/EA2A5822-43D4-4D8D-A18A-B51577A8284C/Library/Application%20Support, "NSFilePath": /var/mobile/Containers/Data/Application/EA2A5822-43D4-4D8D-A18A-B51577A8284C/Library/Application Support]

mbrandonw:
Yes, so then it seems to me this is the problem:

https://github.com/pointfreeco/sqlite-data/blob/65502acdb033ec8025a0bcc443abf2f4ca0598f9/Sources/SQLiteData/CloudKit/SyncEngine.swift#L2242-L2245

We are actively creating the application support directory when establishing the metadatabase. I don't think that really needs to be there and may have just been left over from some other change. It seems we could just remove that to fix this.

But either way, I would ask you to please investigate a better fix than what this PR does right now, because as far as I can tell it should not be necessary to customize where the metadatabase is saved. It is saved in the same directory as the user's database, and so as long as the user is saving their database is the correct place, the metadatbase should be fine.

byMohamedali:
You’re right, the `metadatabaseURL` override was too broad.

  I reworked the fix so we don’t change the public API or thread a new parameter through the
  constructors. Instead, the workaround is scoped to tvOS only: we skip the unconditional
  `Application Support` directory setup in the two internal metadatabase code paths there,
  while preserving the existing behavior on the other platforms.

  So this keeps the old behavior everywhere else and limits the fix to the platform that was
  actually having trouble.

mbrandonw:
I am going to close this because I believe it was fixed in another PR. Please do open a discussion if you are having problems still.

PR #441, opened 2026-04-06
Add missing package dependencies to resolve non-SPM linker failures

scogeo:
## Summary

Fixes linker issues when non-SPM targets depend on `SqliteData` by explicitly declaring missing package dependencies.

## Details

When linking non-SPM targets to `SqliteData`, builds can fail due to undeclared dependencies in the `sqlite-data` package:

- `StructuredQueriesCore` is directly imported in multiple source files but is not listed in the package manifest.
- `StructuredQueriesSQLiteCore` is not explicitly imported, but is resolved transitively by Swift Package Manager. However, this transitive resolution does not carry over to non-SPM targets, resulting in missing symbol linker errors (especially in test builds).

A typical failing dependency chain:

    (.xctest, .xcframework, .appex)
        → (internal SPM library)
            → (sqlite-data)

In these cases:
- Compilation succeeds and dependencies appear resolved.
- Linking fails in test builds or produces warnings in non-test builds.

## Environment

- Xcode 26.4 (17E192)  
- swift-driver 1.148.6  
- Apple Swift 6.3 (swiftlang-6.3.0.123.5 clang-2100.0.123.102)  
- Target: arm64-apple-macosx26.0  

## Resolution

This PR explicitly adds the missing dependencies to the package manifest, ensuring consistent linking behavior across both SPM and non-SPM targets.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -58,6 +58,7 @@ let package = Package(
         .product(name: "Sharing", package: "swift-sharing"),
         .product(name: "StructuredQueriesCore", package: "swift-structured-queries"),
         .product(name: "StructuredQueriesSQLite", package: "swift-structured-queries"),
+        .product(name: "StructuredQueriesSQLiteCore", package: "swift-structured-queries"),
         .product(
           name: "Tagged",
           package: "swift-tagged",
--- a/Package.swift
+++ b/Package.swift
@@ -56,6 +56,7 @@ let package = Package(
         .product(name: "OrderedCollections", package: "swift-collections"),
         .product(name: "Perception", package: "swift-perception"),
         .product(name: "Sharing", package: "swift-sharing"),
+        .product(name: "StructuredQueriesCore", package: "swift-structured-queries"),
         .product(name: "StructuredQueriesSQLite", package: "swift-structured-queries"),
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @scogeo, sorry we never responded to this. We've been modernizing our packages to turn on explicit imports (#467) and so hopefully some of that work will fix any problems you are seeing.

But I am curious about how exactly to reproduce the problem you are running into here. Because StructuredQueriesSQLite re-exports those two libraries, so it should be sufficient to just depend on the one. We'd be happy to merge this if it fixes something, but we just want to understand what is going on.

scogeo:
@mbrandonw No worries. My project is fairly complex, so I would have to create a much simpler project that can reproduce it.  I am hoping it may also just be an Xcode tooling issue that self resolves at some point. If you prefer, we can close this PR for now, and I can open an issue when I have a simpler reproducible project along with a PR that resolves it. 

If I confirm that #467 fixes the issue I will add a followup comment.

scogeo:
@mbrandonw I have verified that #467 has resolved the linker issues I was experiencing , so I'm closing this PR.

PR #455, opened 2026-04-29
Add FetchKeyCache.invalidate for swappable database wrappers.

no33jou:
Linked discussion: https://github.com/pointfreeco/sqlite-data/discussions/454

## Problem

A `DatabaseWriter` wrapper that keeps its own object identity stable across inner-connection swaps causes `@FetchOne` push to stop. `FetchKeyID` uses `ObjectIdentifier(database)`; swift-sharing's persistent-reference cache hands back the reference whose `ValueObservation` is bound to the closed previous connection. `@FetchAll` only escapes by accident (different generic-cast path inside `PersistentReferences`).

Repro (5 sec, runs `swift run Repro` / `swift run Repro -- --fix`):
https://github.com/no33jou/fetchone-mutable-db-repro

## Change

- New file `Sources/SQLiteData/FetchKeyCache.swift` — `public enum FetchKeyCache { public static func invalidate(database:) }`. Internally an `_FetchKeyRegistry` (lock-isolated) keys reattach closures by `ObjectIdentifier(database)`.
- `Sources/SQLiteData/Internal/FetchKey.swift` — `subscribe(...)` now registers a reattach closure on each subscribe; `invalidate` cancels the existing GRDB cancellable and starts a fresh one against the wrapper's current inner connection. The `SharedSubscription`'s cancel closure unregisters from the registry.

`invalidate` does not emit a new value to existing observers. Callers refresh state via the standard `.load(...)` flow after invalidating; this is the same call shape they already use after a connection swap, and it keeps the reattach safely outside of any in-progress observer callback (see code comments for the deadlock scenario this avoids).

Additive: no public API surface change for callers that don't swap inner connections. No behavior change unless `FetchKeyCache.invalidate(database:)` is called.

## Tests

`Tests/SQLiteDataTests/FetchKeyCacheTests.swift` adds two cases against a minimal `SwappableWriter`:

1. `pushDropsAfterInnerSwap_withoutInvalidate` — without `invalidate`, writes to the new connection do not reach `@FetchOne`. Locks the regression in.
2. `pushRestoredAfterInvalidate` — with `invalidate`, push resumes against the new connection.

```sh
swift test --filter FetchKeyCacheTests   # 2/2 pass
swift test                               # 236/236 pass (no regressions)
```

I also verified the tests catch the bug: temporarily reverting `FetchKey.swift` while keeping the test file and `FetchKeyCache.swift` makes test (2) fail with `unread → 0 ≠ 1`. Test (1) keeps passing — it's the regression sanity, not the fix verification.

## Compatibility

- Additive `public enum FetchKeyCache.invalidate(database:)`.
- Internal `_FetchKeyRegistry` is `@unchecked Sendable` guarded by `LockIsolated`. Process-wide lifetime, matching the swift-sharing cache it shadows.
- No change to the existing `FetchKey.subscribe` return value or the public `FetchKeyRequest` protocol.

## Open questions for review

- Naming: `FetchKeyCache.invalidate(database:)` vs `SQLiteData.invalidateFetchKeys(for:)` vs something nested under `FetchKey`. Happy to follow your preference.
- Whether to add a brief `<doc:SwappableDatabase>` article in `Documentation.docc` — can do as a follow-up if useful.

### DIFF
--- a/Sources/SQLiteData/Internal/FetchKey.swift
+++ b/Sources/SQLiteData/Internal/FetchKey.swift
@@ -170,7 +170,7 @@ struct FetchKey<Value: Sendable>: SharedReaderKey {
 }
 
 struct FetchKeyID: Hashable {
-  fileprivate let databaseID: ObjectIdentifier
+  fileprivate let databaseID: AnyHashable
   fileprivate let request: AnyHashableSendable
   fileprivate let requestTypeID: ObjectIdentifier
   fileprivate let scheduler: AnyHashableSendable?
@@ -180,7 +180,11 @@ struct FetchKeyID: Hashable {
     request: some FetchKeyRequest,
     scheduler: (any ValueObservationScheduler & Hashable)?
   ) {
-    self.databaseID = ObjectIdentifier(database)
+    if let identified = database as? PersistentDatabaseIdentity {
+      self.databaseID = identified.persistentIdentity
+    } else {
+      self.databaseID = AnyHashable(ObjectIdentifier(database))
+    }
     self.request = AnyHashableSendable(request)
     self.requestTypeID = ObjectIdentifier(type(of: request))
     self.scheduler = scheduler.map { AnyHashableSendable($0) }
--- a/Sources/SQLiteData/PersistentDatabaseIdentity.swift
+++ b/Sources/SQLiteData/PersistentDatabaseIdentity.swift
@@ -1,0 +1,34 @@
+import Foundation
+import GRDB
+
+/// A hook that lets transparent `DatabaseReader` wrappers participate in
+/// `FetchKey`'s persistent-reference cache.
+///
+/// `FetchKey` keys cached references in part by an identifier derived from the
+/// `DatabaseReader`. By default that identifier is `ObjectIdentifier(database)`,
+/// which is stable for plain `DatabasePool` / `DatabaseQueue`. A wrapper that
+/// swaps its inner pool while keeping its own object identity stable (e.g. an
+/// account-switch wrapper) breaks this assumption: the cached reference's
+/// `ValueObservation` stays bound to the previous inner pool, and writes
+/// against the new pool never reach `@FetchOne` / `@FetchAll` observers.
+///
+/// Wrappers that swap their inner reader can conform to this protocol and
+/// forward `persistentIdentity` to whatever uniquely identifies the *current*
+/// inner connection. After a swap the value must differ from the value before
+/// the swap, so the cache invalidates and the next `.load` rebuilds the
+/// subscription against the new connection.
+///
+/// ```swift
+/// extension AccountDatabase: PersistentDatabaseIdentity {
+///   var persistentIdentity: AnyHashable {
+///     AnyHashable(ObjectIdentifier(currentInnerPool))
+///   }
+/// }
+/// ```
+public protocol PersistentDatabaseIdentity {
+  /// A stable, hashable identity for the current underlying connection.
+  ///
+  /// Two reads against the same conceptual connection return equal values; a
+  /// swap of the underlying connection produces a different value.
+  var persistentIdentity: AnyHashable { get }
+}
--- a/Tests/SQLiteDataTests/PersistentDatabaseIdentityTests.swift
+++ b/Tests/SQLiteDataTests/PersistentDatabaseIdentityTests.swift
@@ -1,0 +1,202 @@
+import ConcurrencyExtras
+import Dependencies
+import DependenciesTestSupport
+import Foundation
+import GRDB
+import SQLiteData
+import StructuredQueries
+import Testing
+
+@Suite("PersistentDatabaseIdentity")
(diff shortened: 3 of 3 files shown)


mbrandonw:
Hi @no33jou, it seems to me that you are only able to reproduce this "bug" by making a custom conformance to `DatabaseReader` and `DatabaseWriter`, which GRDB's docs specifically say is [not valid](https://github.com/groue/GRDB.swift/blob/36e30a6f1ef10e4194f6af0cff90888526f0c115/GRDB/Core/DatabaseReader.swift#L8-L9) to do.

Further, there are other ways to accomplish this "swappable" database idea that are not literally changing the database connection. You can [attach](https://sqlite.org/lang_attach.html)/[detach](https://sqlite.org/lang_detach.html) databases too. I would encourage you to explore that option as an alternative, or you will need to open a discussion on GRDB to get a better understanding of why custom conformances to `DatabaseReader`/`DatabaseWriter` are not allowed.

no33jou:
Thanks for the response. One thing the contract argument doesn't explain: `@FetchAll` and `@FetchOne` diverge on the same wrapper, same swap, same write. Probe on `_add(observation:)` after `mutable.database = poolB`:

```
[_add] Reducer.Value=Result<[Item], Error>  poolPath=B   ← FetchAll re-subscribes
(no _add for FetchOne)                                   ← FetchOne reuses stale ref
```

Why they diverge — swift-sharing init/load asymmetry:

```swift
// SharedReader.init(wrappedValue:_:)         — unwraps .Default
self.init(wrappedValue: wrappedValue(), key.base)

// SharedReader.load(_:)                      — does not unwrap
projectedValue = SharedReader(
  reference: persistentReferences.value(forKey: key, ...)
)

// PersistentReferences.value(forKey:)        — matches by exact type
storage[key.id] as? Weak<Key>
```

Combined with sqlite-data's overload choice (`.Default` for collections, none for single-value), FetchAll lands in different cache buckets at init vs load → miss → reattach. FetchOne lands in the same bucket → hit → stale. FetchAll "working" is a side-effect, not a feature.

Fix (PR ready, ~30 lines + tests, 236 tests pass, no API change for `DatabasePool` / `DatabaseQueue` users):

```swift
public protocol PersistentDatabaseIdentity {
  var persistentIdentity: AnyHashable { get }
}

// FetchKeyID derives databaseID from this hook when available,
// else falls back to ObjectIdentifier(database).
// Wrappers forward to current inner reader → swap invalidates cache.
```

Decoupled from swift-sharing's asymmetry, so future symmetrization won't regress FetchAll.

Branch: https://github.com/no33jou/sqlite-data/tree/fix/fetchkey-id-from-inner-pool
Repro: standalone Swift package, three modes (`bug` / `reset` / `protocol`) demonstrating the divergence and the fix — happy to share on request.

(ATTACH/DETACH is a non-starter for per-user DB isolation — multi-week rearchitecture of StructuredQueries + region tracking. But that's secondary; the divergence reproduces regardless of swap implementation.)

stephencelis:
@no33jou Your branch still depends on [subclassing `DatabaseReader`/`DatabaseWriter`](https://github.com/pointfreeco/sqlite-data/compare/main...no33jou:sqlite-data:fix/fetchkey-id-from-inner-pool#diff-c6ad8ba929e93a8dab6867e388f730fe5e3444429c6de91b5541c2e5e7ecfc50R121), which Brandon mentioned above is not a GRDB-supported thing to do.

rcarver:
fwiw I'm using `ATTACH` to support on-disk file moves against the connected database (document-based app using sqlite in document package). It's a bit tricky but works great. Can package up some code if that's helpful.

dreymonde:
> fwiw I'm using `ATTACH` to support on-disk file moves against the connected database (document-based app using sqlite in document package). It's a bit tricky but works great. Can package up some code if that's helpful.

Would love to take a look if you're up to making a gist or an example repo! Curious about it

no33jou:
Going with a different approach and closing this.

Final direction: per-session `DatabasePool` + SwiftUI tree rebuild on user switch. One pool per session, never swap. `.id(user.id)` on the root view tears down the previous tree before a new pool is observed. Stays entirely within documented sqlite-data usage — no wrapper, no fork. Three tests covering logout→login, closing previous pool, and sequential sessions on the same path all pass.

One takeaway, separate from the wrapper question: `@FetchOne` and `@FetchAll` diverging on the same `defaultDatabase` change was hard to debug. `FetchAll` keeps working in the same view, so you spend a long time looking for what's wrong with your `FetchOne` before realizing it's a framework-level asymmetry (`SharedReader.init` unwraps `.Default`, `load` doesn't; sqlite-data has the overload for collections but not single-value). At minimum a doc note that they can diverge under non-stable-identity scenarios would have saved a couple of debugging days.

Thanks for the discussion.

mbrandonw:
Is this really worth documenting if the only way to run into the problem is to conform to a protocol that the docs explicitly say not to conform to?

PR #464, opened 2026-05-22
Add Android GRDB SQLCipher support

woodymelling:
https://github.com/groue/GRDB.swift/pull/1708

This is a PR to use the same SQLCipher trait that is employed over in GRDB to support Android. 

This is an initial path that seems to allow compiling for Android through skip, but there may be better ways to do it.

It seems that the grdb-sqlcipher needs some updates to match GRDB, so there's a little bit of shim code in here for the time being. I'll work on getting that updated while we discuss

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,11 +24,22 @@ let package = Package(
     .trait(
       name: "SQLiteDataTagged",
       description: "Introduce SQLiteData conformances to the swift-tagged package."
+    ),
+    .trait(
+      name: "GRDBCIPHER",
+      description: "Use the SQLCipher-backed GRDB package."
     )
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-collections", from: "1.0.0"),
-    .package(url: "https://github.com/groue/GRDB.swift", from: "7.6.0"),
+    .package(
+      url: "https://github.com/swift-everywhere/grdb-sqlcipher.git",
+      from: "7.5.0",
+      traits: [
+        .trait(name: "GRDBCIPHER", condition: .when(traits: ["GRDBCIPHER"]))
+      ]
+    ),
+    .package(url: "https://github.com/skiptools/swift-sqlcipher.git", from: "1.3.0"),
     .package(url: "https://github.com/pointfreeco/swift-concurrency-extras", from: "1.0.0"),
     .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.3.3"),
     .package(url: "https://github.com/pointfreeco/swift-dependencies", from: "1.9.0"),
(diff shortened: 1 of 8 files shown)


Sources/SQLiteData/CloudKit/Internal/UserDatabase.swift
```diff
@@ -16,7 +16,7 @@
     }
 
     package func write<T: Sendable>(
-      _ updates: @Sendable (Database) throws -> T
+      _ updates: @escaping @Sendable (Database) throws -> T
```
mbrandonw, line 19: Can you explain why this is necessary? The `database.write` method in GRDB being called below only requires a `@Sendable` closure.

Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -756,15 +756,15 @@
       try tearDownSyncEngine()
       await withErrorReporting(.sqliteDataCloudKitFailure) {
         try await userDatabase.write { db in
-          for table in tables {
+          for table in self.tables {
```
mbrandonw, line 759: Is this because Skip/Android is using an older version of Swift that doesn't allow an implicit `self`? Or something else?

Package.swift
```diff
@@ -24,11 +24,22 @@ let package = Package(
...
       description: "Introduce SQLiteData conformances to the swift-tagged package."
+    ),
+    .trait(
+      name: "GRDBCIPHER",
+      description: "Use the SQLCipher-backed GRDB package."
     )
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-collections", from: "1.0.0"),
-    .package(url: "https://github.com/groue/GRDB.swift", from: "7.6.0"),
```
mbrandonw, line 31: As far as I can tell this is completely removing GRDB proper and using only the SQLCipher version? I don't think we would want to do that for people that don't need SQLCipher.

Package.swift
```diff
@@ -91,6 +107,7 @@ let package = Package(
 )
 
 let swiftSettings: [SwiftSetting] = [
+  .define("GRDBCIPHER", .when(traits: ["GRDBCIPHER"])),
```
mbrandonw: This shouldn't be necessary. Traits already define a flag in Swift that can be used with `#if`.

Package.swift
woodymelling: Oops missed a commit, pushed the change with GRDB as a default trait

Package.swift
woodymelling: Addressed

Sources/SQLiteData/CloudKit/Internal/UserDatabase.swift
woodymelling: This is a req of the current `swift-everywhere/grdb-sqlcipher` fork.

I'm looking into why that's happening, and if it's a straightforward update. 
I'll mark this as draft until we're able to squash this back to the GRDB shape.

Sources/SQLiteData/CloudKit/SyncEngine.swift
woodymelling: No, this is fallout from making this escaping. 
Hopefully the final patch will just be the import changes

PR #466, opened 2026-05-27
Use new task local trait from ConcurrencyExtras.

mbrandonw:
We have a new `.taskLocal` test trait in ConcurrencyExtras (see [here](https://github.com/pointfreeco/swift-concurrency-extras/pull/65)), and so let's start using it in this library. We are able to delete 180 lines of boilerplate for a bunch of task locals we use in our test suite to configure various parts of the sync engine.


Tests/SQLiteDataTests/Internal/TestScopes.swift
```diff
@@ -3,176 +3,30 @@
...
-      try await Self.$prepareDatabase.withValue(prepareDatabase) {
-        try await function()
-      }
-    }
+  var prepareDatabase: @Sendable (UserDatabase) async throws -> Void {
+    _$prepareDatabase.get()
   }
+  let _$prepareDatabase = TaskLocal<@Sendable (UserDatabase) async throws -> Void>(
+    wrappedValue: { _ in }
+  )
```
mbrandonw: Unfortunately the test suite defines the task locals and so we cannot reference the macro generated code (`$prepareDatabase`) from this target. So I've decided to expand the macro code, but alternatively we could have extracted the task locals to their own package.

stephencelis (approved):

PR #470, opened 2026-06-02
Fix error when creating databases on tvOS devices

joseph-elmallah:
Remove redundant `createDirectory` calls that are now handled by `FileManager.default.url(create:true)`.

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -13,10 +13,6 @@
       open "\(url.path(percentEncoded: false))"
       """
     )
-    try FileManager.default.createDirectory(
-      at: .applicationSupportDirectory,
-      withIntermediateDirectories: true
-    )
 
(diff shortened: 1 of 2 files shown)


mbrandonw:
> On tvOS, `.applicationSupportDirectory` is unavailable, so switch to `.cachesDirectory` for the default database path.

I don't think that is the correct move. That will cause people to unknowingly store their database in a place that can be deleted at any moment. Please revert that change. People who develop for tvOS will just have to know that they need to provide a URL to `defaultDatabase` and cannot rely on defaults.

joseph-elmallah:
@mbrandonw I reverted that change and updated the PR description + title to reflect the changes.

mbrandonw (approved):
Thanks, looks good! Did you happen to run the demos and confirm that all still works in the iOS simulator?

joseph-elmallah:
@mbrandonw yes I ran the Demo on a simulator no problem. I also verified the fix does address the problem on a real apple TV. Thanks a lot for your support.

PR #477, opened 2026-06-09
Work around `@_exported`-`public import` bug

stephencelis:
We've relied on fine-grained GRDB exports via, e.g.:

```swift
@_exported import class GRDB.Database
```

But it turns out this breaks as soon as you have an explicit `public import` in the same module:

```swift
public import class GRDB.Database
```

To work around this bug, we can use `#if` blocks to make each mutually exclusive.

### DIFF
--- a/Examples/Reminders/RemindersDetail.swift
+++ b/Examples/Reminders/RemindersDetail.swift
@@ -88,7 +88,7 @@ class RemindersDetailModel: HashableObject {
     }
   }
 
-  private var remindersQuery: some StructuredQueriesCore.Statement<Row> {
+  private var remindersQuery: some Statement<Row> & Sendable {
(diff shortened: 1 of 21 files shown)


mbrandonw (approved):

Examples/Reminders/RemindersDetail.swift
```diff
@@ -88,7 +88,7 @@ class RemindersDetailModel: HashableObject {
     }
   }
 
-  private var remindersQuery: some StructuredQueriesCore.Statement<Row> {
+  private var remindersQuery: some Statement<Row> & Sendable {
```
mbrandonw, line 91: can we drop `Sendable` once we point to newest IssueReporting and Sharing?

PR #500, opened 2026-07-16
Add SyncEngine.pendingChangeCounts

gigq:
This makes the pending save, delete, and database change counts available for things like debug and diagnostics views. It returns nil when the sync engine isn’t running and updates when CKSyncEngine reports a state change.

Tested this in my app, count changes come in live as expected.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -19,6 +19,19 @@
     import UIKit
   #endif
 
+  /// Counts of changes waiting to be sent to CloudKit.
+  @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+  public struct PendingChangeCounts: Equatable, Sendable {
+    /// The number of records waiting to be saved.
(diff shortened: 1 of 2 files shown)


Sources/SQLiteData/CloudKit/SyncEngine.swift
```diff
@@ -422,44 +436,81 @@
       isSendingChanges || isFetchingChanges
     }
 
+    /// Counts of changes waiting to be sent to CloudKit.
+    ///
+    /// This value is `nil` when the sync engine is not running. It is observable, and so accessing
+    /// it from a SwiftUI view will cause the view to update as the pending changes change.
+    public var pendingChangeCounts: PendingChangeCounts? {
```
mbrandonw: Does this really need to be optional? You already know if the engine is running or not via the public `isRunning` property, and so shouldn't this property always reflect the pending changes, regardless of running or not?

gigq: Yeah we can save the pending change count before it's stopped and return those, lets me update the PR.

PR #502, opened 2026-07-17
Propagate suspension handling to CloudKit metadata

phil1995:
## Motivation

`SyncEngine` already includes iOS-specific lifecycle handling: it observes `UIApplication.willResignActiveNotification`, begins a background task, and attempts to send pending private and shared CloudKit changes before the app is suspended.

watchOS does not use this integration and can suspend apps much more aggressively. GRDB supports this environment through the experimental flag `observesSuspensionNotifications`, allowing clients to release database locks when posting `Database.suspendNotification`.

However, while the user database can opt into this behavior, `SyncEngine` maintains a separate CloudKit metadata connection that does not currently participate. This can leave the metadata database holding a SQLite lock during suspension, resulting in a `RUNNINGBOARD 0xdead10cc` termination on watchOS.

### DIFF
--- a/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
+++ b/Sources/SQLiteData/CloudKit/Internal/Metadatabase.swift
@@ -8,7 +8,8 @@
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
   func defaultMetadatabase(
     logger: Logger,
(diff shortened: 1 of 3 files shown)


mbrandonw (approved):
Thanks @phil1995! I made a few small updates, but overall looks good!

Issue #356, opened 2026-01-05
Inconsistent conflict resolution between production and tests


lukaskubanek:
### Description

I’ve been trying to wrap my head around the implementation of the built-in “field-wise last edit wins” strategy, and I’m honestly having a hard time. To better understand what’s going on, I started documenting the current behavior in [this document](https://github.com/structuredpath/sqlite-data-sync-notes/blob/main/CurrentBehavior.md). While I still don’t have all the answers, I ran into some unexpected behavior that felt worth reporting, as it might help clarify the intended design.

It traces back to the [purpose of the `didSet` flag](https://github.com/pointfreeco/sqlite-data/discussions/272#discussioncomment-15059479) in `CKRecord.update(with:row:columnNames:parentForeignKey:)`, which is used during upserts. As I understand it, this logic attempts to restore values from the _last-known server record_ onto an incoming server record. In my mental model, this should be superfluous (and really a no-op), as I was assuming the last-known server record represents the most recent record that was either accepted by or fetched from the server. In the three-way merge terms, this would be the _ancestor_.

However, while digging deeper, I found two spots in the codebase where a record is saved as the last-known server record _before_ it has been accepted by the server. I went into more detail on these in the _warning_ sections of the [linked document](https://github.com/structuredpath/sqlite-data-sync-notes/blob/main/CurrentBehavior.md). In particular, the [premature save during send](https://github.com/pointfreeco/sqlite-data/blob/df360b234eec719360b1fa7912efeff54d6eee8e/Sources/SQLiteData/CloudKit/SyncEngine.swift#L1174) looked worrying to me, as it seems to break the mental model of the last-known server record.

When testing this with the sample apps, I noticed that the record was _not_ actually being saved during send. Inspecting `SyncEngine.refreshLastKnownServerRecord(_:)` revealed that this was due to  the check of the `modificationDate` between the saved last-known server record and the given record, which happened to be equal, as the given record was based off of the last-known server record.

In tests, however, the record _is_ being saved. The reason appears to be that the mock server does not populate `modificationDate` on records it reports back to the sync engine. While `CKRecord.modificationDate` can’t be set directly, I was able to swizzle it and have the mock server simulate a save timestamp. With that change in place, behavior became closer to production but two `MergeConflictTests` started producing different results, as they had previously been operating on a polluted last-known server record:

[...]

Issue #274, opened 2025-10-24
Data divergence between devices after concurrent offline edits


lukaskubanek:
### Description

While experimenting with the library in preparation for #272, I ran into a case where a concurrently edited field diverged between devices.

Let’s consider a single `TEXT` column of a table (I was using `Reminder.title` in the Reminders example app) and two devices, A and B, editing the field of the same row concurrently. To simulate a real-world scenario, device A is put offline during the editing phase.

### Scenario 1

In the first scenario, device B makes its edit before device A. Everything works as expected. When device A comes online, it’s prevented from sending its value until it downloads device B’s update. Since device A’s change is considered the last write, it’s then correctly propagated to device B.

| Device A | Device B |
| -------- | -------- |
| `foo`    | `foo`    |
| offline  |          |
|          | `fooB`   |
|          | send     |
| `fooA`   |          |
| online   |          |
| send 💥   |          |
| fetch    |          |
| send     |          |
|          | fetch    |
|          | `fooA`   |

### Scenario 2

In the second scenario, the steps are similar, only the order of edits is swapped. When device A comes online, it also fetches device B’s value. But this time, since device B’s edit is the last write, device A should adopt that value and drop its local change. However, this doesn’t happen. Device A keeps its local value, which causes the SQLite database to diverge from the CloudKit record, as the local value correctly isn’t synced with CloudKit anymore.

| Device A | Device B |
| -------- | -------- |
| `foo`    | `foo`    |
| offline  |          |
| `fooA`   |          |
|          | `fooB`   |
|          | send     |
| online   |          |
| send 💥   |          |
| fetch ⚠️ |          |
| send ⚠️  |          |
|          | fetch    |

It seems that device A’s local value is still considered a modification during the fetch. The expected behavior would be for the logic to recognize that the local change is outdated and let it be overwritten by the server value.

This might be better addressed with an abstraction built as part of #272, but I think it’s better to tackle it in piecemeal steps.

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] I have determined whether this bug is also reproducible in a vanilla GRDB project.
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/sqlite-data/issues) or [discussion](https://github.com/pointfreeco/sqlite-data/discussions).

### SQLiteData version information

`main`

### Destination operating system

[...]

PR #506, opened 2026-07-23
Fix permanent sync divergence when userModificationTime stamps tie

notcome:
## Summary

Two devices writing the same field with identical microsecond `userModificationTime` stamps can diverge **permanently and silently**: one device keeps its own value forever while the server and all other devices converge on a different value, and the stuck device reports fully synced. We hit this reproducibly (8/9 scripted trials) with near-simultaneous same-primary-key inserts from two processes — concurrent writers land nanosecond-identical stamps often, since the clock has microsecond resolution and timers coalesce.

## Mechanism

Two behaviors compound:

1. **Self-preferring tie-break.** The merge guards in `CloudKit+StructuredQueries.swift` (`setValue`/`setAsset`/`removeValue`) use `<=`, so on a per-field stamp tie the local/base value overwrites the incoming committed server value. Self-preference is not a total order: both replicas of a tie each conclude "mine wins" and resolve the same tie oppositely.
2. **Override without enqueue on the fetch path.** When `handleFetchedRecordZoneChanges` → `upsertFromServerRecord` preserves local values over an incoming committed record, no upload is enqueued — the `.serverRecordChanged` conflict path re-enqueues (`SyncEngine.swift`), but the fetch path does not.

End to end: A and B insert the same key with stamp *t*. A commits first. B's save conflicts; the tie keeps B's value; the conflict path re-enqueues; B's retry commits — B and the server agree. A then *fetches* B's commit: the tie keeps A's value, the merge-mutated record (server change tag around A's values) is stored as A's base, and nothing is enqueued. A's row now differs from its base on every future delivery, so the `isRowValueModified` branch discards every subsequent server value while advancing the base — a silent absorbing state.

## The fix

1. **Strict `<` for inbound merges.** The three setters gain a `requireStrictlyNewer` parameter (default `false`, so outgoing record construction — where an equal-stamp local edit must still apply — is unchanged). The inbound merge passes `true`: a base field overrules an incoming committed value only when strictly newer. On a tie the committed value wins. CloudKit's CAS chain already totally orders each record's history and replays it identically to every replica, so commit order serves as a deterministic tie-breaker shared by all replicas — no schema change, and safe in a mixed fleet of patched and unpatched clients (a patched device defers on ties; an unpatched peer's conflict path pushes its value; everyone converges on it).
2. **Enqueue on override.** When an inbound merge on the fetch path actively overrules incoming committed field values (base field strictly newer — e.g. clock skew between writers), a `.saveRecord` is enqueued for that record, mirroring the conflict path, so the surviving value is re-asserted to the server and replicas that never saw it converge. Two deliberate scope limits, both shaped by the existing test suite: stale replays (incoming older than the stored base) are suppressed, since a replay legitimately loses and must not cause churn; and row-level preservation of pending local edits (`isRowValueModified`) does not enqueue, since those uploads are already owned by the change triggers.

## Tests

Two regression tests in `MergeConflictTests` using the existing mocks:

- `equalTimestampConflictConvergesToServerValue`: a fetched record carrying a different value at an equal field stamp must be adopted, leaving row, server, and base converged.
- `olderTimestampFetchReassertsLocalWinToServer`: a later commit carrying an older field stamp loses the merge, and the surviving local value is enqueued and pushed so the server converges on the LWW winner.

Full suite passes (241 tests). With `Sources/` reverted to `main` and the tests kept, both fail with the expected signatures (device keeps `"Mine"` over committed `"Theirs"`; the `.saveRecord` never enqueued).

One observation for maintainers, made while validating the red case: `MockCloudDatabase.modifyRecords` stores and returns the same `CKRecord` instance (`MockCloudDatabase.swift:204-205`), so a merge that mutates a fetched record in place also mutates mock server storage — under the buggy code the mock "server" appears to adopt the stale device's value when real CloudKit would not. Happy to file that separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
+++ b/Sources/SQLiteData/CloudKit/CloudKit+StructuredQueries.swift
@@ -156,10 +156,13 @@
     package func setValue(
       _ newValue: some CKRecordValueProtocol & Equatable,
       forKey key: CKRecord.FieldKey,
-      at userModificationTime: Int64
+      at userModificationTime: Int64,
+      requireStrictlyNewer: Bool = false
     ) -> Bool {
       guard
-        encryptedValues[at: key] <= userModificationTime,
+        requireStrictlyNewer
+          ? encryptedValues[at: key] < userModificationTime
+          : encryptedValues[at: key] <= userModificationTime,
         encryptedValues[key] != newValue
       else { return false }
       encryptedValues[key] = newValue
@@ -172,7 +175,8 @@
     package func setAsset(
       _ newValue: CKAsset,
       forKey key: CKRecord.FieldKey,
-      at userModificationTime: Int64
+      at userModificationTime: Int64,
+      requireStrictlyNewer: Bool = false
     ) -> Bool {
       @Dependency(\.dataManager) var dataManager
       guard
@@ -180,7 +184,9 @@
         let hash = dataManager.sha256(of: fileURL)
       else { return false }
       guard
-        encryptedValues[at: key] <= userModificationTime,
+        requireStrictlyNewer
+          ? encryptedValues[at: key] < userModificationTime
+          : encryptedValues[at: key] <= userModificationTime,
         encryptedValues[hash: key] != hash
       else { return false }
 
@@ -223,9 +229,13 @@
     @discardableResult
     package func removeValue(
       forKey key: CKRecord.FieldKey,
-      at userModificationTime: Int64
+      at userModificationTime: Int64,
+      requireStrictlyNewer: Bool = false
     ) -> Bool {
-      guard encryptedValues[at: key] <= userModificationTime
+      guard
+        requireStrictlyNewer
+          ? encryptedValues[at: key] < userModificationTime
+          : encryptedValues[at: key] <= userModificationTime
       else {
         return false
       }
@@ -283,26 +293,42 @@
       }
     }
 
+    @discardableResult
     func update<T: PrimaryKeyedTable>(
       with other: CKRecord,
       row: T,
       columnNames: inout [String],
       parentForeignKey: ForeignKey?
-    ) {
+    ) -> Bool {
       typealias EquatableCKRecordValueProtocol = CKRecordValueProtocol & Equatable
 
       self.userModificationTime = other.userModificationTime
+      var didPreserveLocalValues = false
       for column in T.TableColumns.writableColumns {
         func open<Root, Value>(_ column: some WritableTableColumnExpression<Root, Value>) {
           let key = column.name
           let keyPath = column.keyPath as! KeyPath<T, Value.QueryOutput>
           let didSet: Bool
           if let value = other[key] as? CKAsset {
-            didSet = setAsset(value, forKey: key, at: other.encryptedValues[at: key])
+            didSet = setAsset(
+              value,
+              forKey: key,
+              at: other.encryptedValues[at: key],
+              requireStrictlyNewer: true
+            )
           } else if let value = other.encryptedValues[key] as? any EquatableCKRecordValueProtocol {
-            didSet = setValue(value, forKey: key, at: other.encryptedValues[at: key])
+            didSet = setValue(
+              value,
+              forKey: key,
+              at: other.encryptedValues[at: key],
+              requireStrictlyNewer: true
+            )
           } else if other.encryptedValues[key] == nil {
-            didSet = removeValue(forKey: key, at: other.encryptedValues[at: key])
+            didSet = removeValue(
+              forKey: key,
+              at: other.encryptedValues[at: key],
+              requireStrictlyNewer: true
+            )
           } else {
             didSet = false
           }
@@ -334,13 +360,17 @@
           }
           if didSet || isRowValueModified {
             columnNames.removeAll(where: { $0 == key })
-            if didSet, let parentForeignKey, key == parentForeignKey.from {
-              self.parent = other.parent
+            if didSet {
+              didPreserveLocalValues = true
+              if let parentForeignKey, key == parentForeignKey.from {
+                self.parent = other.parent
+              }
             }
           }
         }
         open(column)
       }
+      return didPreserveLocalValues
     }
 
     package var userModificationTime: Int64 {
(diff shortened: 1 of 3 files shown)


notcome:
Human message: so Fable was doing some automatic integration testing by hand and hit this issue. At its core it is not hard to understand: if two devices modify at the same time, we will have a tie, and our code is essentially local always wins. We should end up with a ping pong situation, but the reconciliation code on the original winner masks it.

The fix is essentially, if values differ and field timestamp coincide, we always treat server record as winner (Change 1). We additionally add an re-enqueue write on conflict from syncing (Change 2). Change 1 is very understandable to me, but the diff is still bigger than I would hope. And Change 2 is arguably a bit over defensive, but it could rescue existing deployment from this bug – if such bug ever materializes, which would be so unlucky.

I tried my best to understand your codebase but this is a very intricate matter, so please take everything other than the idea a huge grain of salt. I did do the PR under a fresher context, if that helps boost your confidence.

mbrandonw:
Hi @notcome, thanks for disclosing your AI use to open this PR and providing a shorter analysis of the situation. But just to confirm, you didn't actually encounter this on real devices, right? It is of course trivial to demonstrate the problem with the test suite because time is controlled, and so easy to perform two saves at the same time, but this happening on two real devices is about as likely as winning the lottery. That doesn't mean this shouldn't be investigated, but I do want to make it clear that any time we put into this is almost certainly chasing a problem that will never happen in reality.

Can you explain more how you ran into the problem in an integration test? 

And as for this PR, I haven't looked at the implementation, but I was interested to see what the tests were trying to show. Unfortunately the tests aren't really testing a real life situation. The first one is emulating a record being updated locally and sync'd to the server, and then a remote client downloads that record and edits it _at the same time_ and sends it back to the server. The chances of that are even less than two devices editing at the same microsecond. And the second test is also doing something illegitimate in that it is directly editing the timestamp in the record rather than going through the `setValue(_:forKey:at:)` helper.

I do think a few (legitimate) tests in the suite with some `withKnownIssue`s to document the behavior would be a good thing. And while investigating this I did find a few problems with the `MockCloudDatabase` that needed to be [fixed](https://github.com/pointfreeco/sqlite-data/pull/507).

mbrandonw:
Hi @notcome, I'm curious to get your thoughts on the above. And also inclined to close this PR as this approach is probably not the one we would take. And as I mentioned above, we would prefer to start with a rock solid test case that shows the actual problem, which the current tests do not (they take shortcuts to force a problem that would not happen in real life).

lukaskubanek:
Hi @mbrandonw and @notcome, chiming in, as I’ve spent some time in this part of the codebase. Here are my observations…

The upsert logic has quite a few issues beyond the one reported here. I started reporting them in #274 and #356, but then decided to holistically document the current behavior of the CloudKit sync in [this document](https://github.com/structuredpath/sqlite-data-sync-notes/blob/main/CurrentBehavior.md). I’ve highlighted several issues in the warning sections, including the equal-timestamp tie in the `<=` comparison that this PR targets (see the warnings in section 2.3).

The most notable problem is that the merge logic runs on every upsert (of an already synced record), gated only on whether a last-known server record and a local row exist, never on whether a logical conflict actually exists.

That is what makes this scenario possible in the first place. The fetching device has no pending changes, so there is nothing to resolve, and the only correct outcome is to apply the incoming server record as is. The tie-break inside the merge is just where the missing distinction becomes observable. A fix at that level treats the symptom rather than the cause.

On the implementation here: change tags are the right tool for detecting server-side changes, but they are opaque values with no defined ordering, so comparing them with `<` is not a legitimate operation. The comparison in this PR also operates on `_recordChangeTag`, a test-only field that only the mock server populates (with an ordered counter, which is the only reason the comparison behaves in tests). In production that field is never set, so this branch is effectively dead code on real devices. The reliable check is a mere inequality of the actual `recordChangeTag` between the last-known server record and the incoming one, which tells whether the server record has changed. The mock infrastructure can’t exercise that yet, since it doesn’t expose its tags through the real `recordChangeTag` property, but I have [that change lined up](https://github.com/structuredpath/sqlite-data/commit/a11ce8a5fabe64f141894c573bb3a84ce866fb31) as a follow-up to PR #411, using the same swizzling approach as for `modificationDate`.

I’ve put a lot of effort into revising this whole area while implementing customizable conflict resolution. Besides other improvements, the logic [explicitly classifies each upsert](https://github.com/structuredpath/sqlite-data/blob/ceb3884f0b63d7bfb75cd3529be2f72861710dac/Sources/SQLiteData/CloudKit/SyncEngine.swift#L2023-L2097) as skip (server unchanged, detected via change-tag equality), apply (no local changes, the server record is applied wholesale), or resolve (a genuine conflict handed to a resolution policy). Under that model the scenario in this PR never reaches a merge, because a device without pending changes simply applies the server record. A working prototype is ready in [my fork](https://github.com/structuredpath/sqlite-data/tree/sp-main) (branch `sp-main`).

I’m ready to extract pieces into focused PRs and submit them. 😉

lukaskubanek:
> I do think a few (legitimate) tests in the suite with some `withKnownIssues` to document the behavior would be a good thing.

See #414.

notcome:
Sorry for this late action! I got a bit carried over in preparing our own app! To answer @mbrandonw your questions:

> But just to confirm, you didn't actually encounter this on real devices, right? It is of course trivial to demonstrate the problem with the test suite because time is controlled, and so easy to perform two saves at the same time, but this happening on two real devices is about as likely as winning the lottery. Can you explain more how you ran into the problem in an integration test?

Yes, we didn't hit it during real devices. We only hit them in a Claude authored end-to-end test. Basically, the setup was to run two processes that modify the same record simultaenously. We had a bash script that launched two Swift processes within microseconds, and then each Swift process chose to sleep for three seconds before doing CloudKit operations. We managed to hit this issue using this setup **8 out of 9 times**. Claude confidently told me that this is due to Darwin choose to coalesce timers, which I don't know if it s true. But in any case, we can get **identical timestamps from different processes on the same Mac** fairly easily.

I totally agree with you that this is just so unlikely to happen. It just itches a little when you know a tie is theoretically possible, :).

@lukaskubanek Your note is amazing and answered some of my own confusion about the situation. Since you have a way more systematic approach to this problem, I would close this PR.

PR #507, opened 2026-07-23
Return copied CKRecords when modifying database.

mbrandonw:
We have a few places we are accidentally returning references from the mock database that exist in the DB's storage. That allows the outside to make edits to the references in the DB. I've put in a few extra `copy()`s to fix this.


Tests/SQLiteDataTests/CloudKitTests/SyncEngineDelegateTests.swift
```diff
@@ -234,7 +234,7 @@
     deinit {
       guard wasCalled.withValue(\.self)
       else {
-        Issue.record("Delegate method 'syncEngine(_:accountChanged:)' was not called.")
+        //Issue.record("Delegate method 'syncEngine(_:accountChanged:)' was not called.")
```
stephencelis:
```suggestion
        Issue.record("Delegate method 'syncEngine(_:accountChanged:)' was not called.")
```

stephencelis (approved):

lukaskubanek:
I didn't have time to check in detail, but it looks like a duplicate of #412, which I've filed a while ago.

mbrandonw:
Hi @lukaskubanek, yeah good catch, these PRs are closely related. Mine chose to fix the problem by forcing the `copy()`s to be in the mock database so that we never return references to stored data (value types would fix this, but I think we're stuck with references). So I think we can close your PR and I've added you as a co-author to this PR (hopefully GH picks up the attribution!). Thanks for investigating the problem and sorry we didn't get to it earlier. 

And I know there's a lot more we need to get back into, so hopefully sometime soon :)

lukaskubanek:
Hi @mbrandonw, I just looked at the diffs and your approach in this PR is way more robust and handles all the cases related to shared `CKRecord` references in the `MockCloudDatabase`. In #412, I only addressed the minimal change for the purpose of getting to customizable conflict resolution. So yeah, this PR supersedes mine, so I’ll go ahead and close it. And thanks for the attribution!

Tests/SQLiteDataTests/CloudKitTests/SyncEngineDelegateTests.swift
lukaskubanek: There’s not much context in this change suggestion, but it’s likely related to #509.

mbrandonw: Yep, thanks for filing that issue. We will look into that soon.

PR #511, opened 2026-07-24
Fix sync delegate test crash in Xcode 27 beta.

### LINKED ISSUE #509: Issue recorded in `MyDelegate.deinit` crashes individual test runs in Xcode 27
## Description

`SyncEngineDelegateTests.accountChanged` injects its delegate through a test trait:

```swift
@Test($syncEngineDelegate.set(MyDelegate()))
func accountChanged() async throws { … }
```

Trait values seem to be constructed during test planning, so the `MyDelegate` instance is created for every run, including runs where `accountChanged` is not included.  In those runs the instance is discarded during planning, its `deinit` finds `wasCalled == false`, and records an issue outside of any running test:

```swift
deinit {
  guard wasCalled.withValue(\.self)
  else {
    Issue.record("Delegate method 'syncEngine(_:accountChanged:)' was not called.")
    return
  }
}
```

## Consequences

In Xcode 27 (beta 3 and beta 4), running any individual test or suite from this package crashes the test runner, while the full suite runs fine. It seems that the crash happens when Xcode processes the issue recorded outside of a test, though it's unclear whether it's meant to handle that at all. Under `swift test` it seems to work fine.

The issue was initially observed while working on #508, although it’s unrelated.

## Reproduction

On `main`, with Xcode 27 beta 3 or 4:

```sh
xcodebuild test -scheme sqlite-data-Package -destination 'platform=macOS' \
  -only-testing:SQLiteDataTests/UserlandTests
```

fails with `Exceeded max restart count of 2. (Underlying Error: Crash: xctest at ABI.EncodedEvent.encode(to:))`.

---

*Disclaimer: AI assistance (Claude Code with Fable model) was used to investigate this issue and prepare this report.*

### Checklist

- [x] I wrote this in my own words, and aimed to be as succinct as possible, even if I used AI to help research its content.
- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] I have determined whether this bug is also reproducible in a vanilla GRDB project.
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/pointfreeco/sqlite-data/issues) or [discussion](https://github.com/pointfreeco/sqlite-data/discussions).

### SQLiteData version information

00c54588254acc47adef69970b72b80b76e5589d

### Xcode version information

Xcode 27 beta 3 & 4

### Swift Compiler version information

```shell
swift-driver version: 1.168.5 Apple Swift version 6.4 (swiftlang-6.4.0.27.1 clang-2100.3.27.1)
Target: arm64-apple-macosx27.0.0
```

mbrandonw:
Fixes #509.

Something different (I think!) in Xcode 27 causes traits to be evaluated on tests that are not actually run. That, coupled with `SyncEngineDelegate`s being kept in memory for too long, caused a crash when running any tests that did not use the sync delegate trait. The fix is just to weakly hold onto the delegate in the `SyncEngine`. 

That does mean that if anyone ever created a sync engine like this:

```swift
SyncEngine(
  …,
  delegate: MyDelegate()
)
```

…will have their delegate immediately deallocated, which would be a bug. Luckily all of our docs and sample apps show off holding on the delegate explicitly, which is standard practice in most Apple APIs.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -34,7 +34,11 @@
     let foreignKeysByTableName: [String: [ForeignKey]]
     package let syncEngines = LockIsolated<SyncEngines>(SyncEngines())
     package let defaultZone: CKRecordZone
-    weak let delegate: (any SyncEngineDelegate)?
+    #if compiler(<6.2.3)
+      weak var delegate: (any SyncEngineDelegate)?
+    #else
+      weak let delegate: (any SyncEngineDelegate)?
+    #endif
     let defaultSyncEngines:
       @Sendable (any DatabaseReader, SyncEngine)
         -> (private: any SyncEngineProtocol, shared: any SyncEngineProtocol)
@@ -944,6 +948,14 @@
       }
     }
   }
+
+  #if compiler(<6.2.3)
+    @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+    extension SyncEngine: @unchecked Sendable {}
+  #else
+    @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
+    extension SyncEngine: Sendable {}
+  #endif
 
   extension PrimaryKeyedTable {
     @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -34,7 +34,7 @@
     let foreignKeysByTableName: [String: [ForeignKey]]
     package let syncEngines = LockIsolated<SyncEngines>(SyncEngines())
     package let defaultZone: CKRecordZone
-    let delegate: (any SyncEngineDelegate)?
+    weak let delegate: (any SyncEngineDelegate)?
     let defaultSyncEngines:
       @Sendable (any DatabaseReader, SyncEngine)
         -> (private: any SyncEngineProtocol, shared: any SyncEngineProtocol)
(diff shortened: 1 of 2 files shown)


stephencelis (approved):

lukaskubanek:
@mbrandonw Thanks for looking into it. I’ve tested it with Xcode 27 β4 and ef3cc191ba1fe1443c71c4eeeebb7d2bf1bbab12. Unfortunately, I’m still running into the crash when executing individual, unrelated tests. The `MyDelegate` instance gets intialized and deinitialized, which leads to the recording of the issue, which in turn crashes the test system.

mbrandonw:
Hi @lukaskubanek, thanks for checking. I was playing with two different fixes at once (the one in this PR, and then https://github.com/pointfreeco/swift-concurrency-extras/pull/74) and maybe due to cached artifacts or pure luck it seemed like only the changes in this PR were necessary. But I think the changes in ConcurrencyExtras is the real fix. Wanna point to that branch locally to see if it fixes for you? (Edit: I just merged the PR so point to `main` instead).

If so, then I think we will close out this PR and just have people point to the newest ConcurrencyExtras if they need to run tests locally.

lukaskubanek:
@mbrandonw I pointed swift-concurrency-extras to `main` and I’m confirming that the patch in pointfreeco/swift-concurrency-extras#74 fixed the issue. So yes, this PR shouldn’t be needed and can be closed. The same goes for #509.

By the way, using `@autoclosure` for the trait parameter was what I had intuitively thought of as well. 🙂

mbrandonw:
Since this turned out to be a fix in just ConcurrencyExtras I am going to close this out.

PR #522, opened 2026-08-11
Re-record snapshots

stephencelis:
StructuredQueries has a bug fix with `IN` subquery rendering that we should pull in.

### DIFF
--- a/Tests/SQLiteDataTests/CloudKitTests/AccountLifecycleTests.swift
+++ b/Tests/SQLiteDataTests/CloudKitTests/AccountLifecycleTests.swift
@@ -40,7 +40,7 @@
       }
 
       @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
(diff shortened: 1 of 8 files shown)


Tests/SQLiteDataTests/CloudKitTests/AccountLifecycleTests.swift
```diff
@@ -40,7 +40,7 @@
       }
 
       @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
-      @Test($accountStatus.set(.noAccount))
```
stephencelis, line 43: `set` was deprecated, so addressing those warnings at the same time.

mbrandonw: yeah, and soon to need `withValue:` argument too...

mbrandonw (approved):

PR #524, opened 2026-08-16
[WIP] feat: add a method to force sync the syncEngine without deleting the local data

MojtabaHs:
Based on the very recent conversation [in Slack](https://pointfreecommunity.slack.com/archives/C08CZUG4Z9D/p1786892634659759?thread_ts=1786879103.115709&cid=C08CZUG4Z9D), I found that tearing down and setting up the sync engine forces it to synchronize changes with CloudKit.

I think the underlying issue is likely related to the sync triggers, but this workaround works for now.

Also, to keep a single source of truth, I reused the existing implementation by adding an argument to the existing function instead of introducing a separate code path.

Please take a look, test, and let me know what you think.

### DIFF
--- a/Sources/SQLiteData/CloudKit/SyncEngine.swift
+++ b/Sources/SQLiteData/CloudKit/SyncEngine.swift
@@ -759,17 +759,26 @@
     /// > Important: It is only appropriate to call this method when the device's iCloud account
     /// > logs out or changes.
     public func deleteLocalData() async throws {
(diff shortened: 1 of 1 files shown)


MojtabaHs:
Update: 
more aggressive testing shows that it can leave triggers dangling and throw:
`SQLite error 1: no such trigger sqlitedata_icloud_after_primary_key_change_on...`

The data also becomes unstable, showing different states after each refresh.

PR #533, opened 2026-08-21
feat: expose recovery actions only for recoverable `SyncEngine.SharingError`s

MojtabaHs:
## Summary

Expose a public recovery action for `SyncEngine.SharingError` without making the error type itself public.

This allows callers to determine when a sharing operation can be recovered by starting the sync engine or sending pending changes and retrying.

### Usage

For example, a sharing operation can use the recovery action to synchronize the engine and retry when possible:

```swift
do {
  sharedRecord = try await syncEngine.share(record: myRecord) { ... }
} catch let error as SyncEngine.RecoverableSharingError {
  switch error.recoveryAction {
  case .sendChangesAndRetry: try await syncEngine.sendChanges()
  case .startAndRetry: try await syncEngine.start()
  case .none: throw error // Re-throws unrecoverable errors
  }
  sharedRecord = try await syncEngine.share(record: myRecord) { ... }
}
```

### Motivation

Following the recent [thread on the sqlite-data Slack channel](https://pointfreecommunity.slack.com/archives/C08CZUG4Z9D/p1787338631450289), it turns out that `SharingError` is currently private, but some of its errors provide actionable recovery guidance in their `debugDescription`.

For example, `.recordMetadataNotFound` explicitly suggests calling `sendChanges()` to force synchronization:
https://github.com/pointfreeco/sqlite-data/blob/f6bd67aaff68cfbbed7f0a575248306a70d3e3ae/Sources/SQLiteData/CloudKit/CloudKitSharing.swift#L138-L143

Without a public recovery API, callers cannot reliably identify this condition and perform the suggested recovery.

This change exposes the recovery contract while keeping the underlying `SharingError` and its internal reasons private.

### Changes

- Add `SyncEngine.SharingErrorRecoveryAction`.
- Add `SyncEngine.RecoverableSharingError` to expose recovery actions without exposing `SharingError`.
- Provide `.startAndRetry` for `.syncEngineNotRunning`.
- Provide `.sendChangesAndRetry` for `.recordMetadataNotFound`.
- Add tests covering the exposed recovery actions following the existing testing pattern and using inline snapshots.

### DIFF
--- a/Sources/SQLiteData/CloudKit/CloudKitSharing.swift
+++ b/Sources/SQLiteData/CloudKit/CloudKitSharing.swift
@@ -33,12 +33,27 @@
 
   @available(iOS 17, macOS 14, tvOS 17, watchOS 10, *)
   extension SyncEngine {
+    /// Describes an action that can be performed by the caller to recover from
+    /// an error and retry the failed operation.
     public enum RecoveryAction {
+      /// No sync metadata found for record. Has the record been saved to the database
+      /// and synchronized to iCloud? Invoke 'SyncEngine.sendChanges()' to force
+      /// synchronization.
       case sendChangesAndRetry
+      /// Sync engine is not running. Make sure engine is running by invoking the 'start()'
+      /// method, or using the 'startImmediately' argument when initializing the engine.
       case startAndRetry
     }
 
+    /// A protocol for errors that provide a system-defined recovery action.
+    ///
+    /// A conforming error may provide a recovery action when the failed
+    /// operation can potentially succeed after performing an additional
+    /// synchronization operation.
     public protocol RecoverableSharingError: LocalizedError {
+      /// The recovery action that can be performed before retrying the
+      /// failed operation, or `nil` when no automatic recovery action
+      /// is available.
       var recoveryAction: RecoveryAction? { get }
     }
   }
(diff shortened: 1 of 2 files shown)


ArchimedesTrajano (approved):
Thank you! I think this can be useful until the error is exposed, and potentially even longer. This covers my needs.

stephencelis:
@MojtabaHs Thanks for exploring! While we definitely want to do this, it's something we also want to take some care to design properly, either determining a finite set of enum cases that is fully extensible in the future for our needs, or losing the closed enum entirely so that we can support more kinds of errors in the future. Going to close this for now but we are tracking the issue internally. Also either of you can open a public issue if you want it tracked more publicly.

ArchimedesTrajano:
Is it possible to merge the latest version into this PR?

MojtabaHs:
@ArchimedesTrajano If you mean [my branch](https://github.com/MojtabaHs/sqlite-data/tree/public-sharing-error), I did. But you should follow [Brandon’s advice](https://pointfreecommunity.slack.com/archives/C08CZUG4Z9D/p1787346206924199?thread_ts=1787338631.450289&cid=C08CZUG4Z9D), or just wait and follow what Stephen suggested.

Also, please fork it and maintain the code yourself. I don’t plan to keep this around.

PR #544, opened 2026-09-06
Disable unused Sharing traits

Econa77:
SQLiteData only uses Sharing's core APIs, but currently enables its default traits. 

Raise the minimum Sharing version to 2.10.1 and specify `traits: []` in the Swift 6.1+ manifests. Also remove an unused `CustomDump` import from `SyncEngineDelegate.swift`.

This allows applications to remove `IdentifiedCollections` from their dependencies when it is not otherwise required, since SQLiteData no longer pulls it in through Sharing.

Although the relevant traits were introduced in 2.10.0, the minimum version is set to 2.10.1 because it [fixes a non-Apple OS build regression introduced in 2.10.0](https://github.com/pointfreeco/swift-sharing/releases/tag/2.10.1).

Applications that need Sharing's optional features can enable the corresponding traits by adding `swift-sharing` as a direct dependency. Existing trait forwarding to StructuredQueries remains unchanged.


mbrandonw (approved):
Thanks for this @Econa77. I made one small change to enable the `CasePaths` trait in Sharing when the corresponding trait is enabled in SQLiteData.

mbrandonw:
Hi again @Econa77, unfortunately due to some SwiftPM bugs we can't merge this right now. As you can see in, we now have failing CI actions and it's due to SwiftPM's inability to resolve the package graph with these changes.

We have run into this before but could only vaguely remember the details, but this is a problem that has been in SwiftPM since traits were introduced. To reproduce you can create a PackageC that conditionally depends on a PackageD based on a trait (defaulted on), and then create a PackageA that depends on PackageC with the trait off, and create a PackageB that depends on PackageC with the default traits. It is currently not possible to then depend on PackageA and PackageB at the same time.

It's a bummer, but in the future once we are ready to go to 2.0 on some of these libraries we will be turning all traits off by default, and in that situation this change would have worked.

mbrandonw:
BTW here is a SwiftPM issue you can track too: https://github.com/swiftlang/swift-package-manager/issues/10397. Looks like it's not going to make Swift 6.4, but if it ever makes it to Swift 6.5 we can consider adding a new Package.swift for disabling the trait.

Econa77:
@mbrandonw Thanks for the detailed explanation and for linking the SwiftPM issue. I understand why this can't be merged right now. Making all traits opt-in in future major releases sounds good to me, as it would help applications avoid pulling in dependencies they don't need.

Since there doesn't seem to be a workaround at the moment, I'll close this PR for now. Thanks for taking the time to review it and look into the issue 🙏 

I'd also like to thank you and the Point-Free team for making these wonderful libraries available. I use them regularly, and they've been a huge help in my projects. I'd love to contribute again when the opportunity comes up.

PR #546, opened 2026-09-11
Update CloudKit sync docs link

DominikGrodl:
The link for CloudKit Synchronization docs currently points to the Extended Modules section, which seems wrong. IMHO it should point to the "Getting started with CloudKit" section.


mbrandonw (approved):
Thanks @DominikGrodl!

